PR #15, opened 2019-05-07
Add multiple config tests

belldandyxtq:
This commit adds multi-config test to test with different python
versions.
This commit tests the latest minor versions of all python > 3.5.


kuenishi:
Try [tox](https://tox.readthedocs.io/en/latest/) rather than running multiple containers from single image w/pyenv.

belldandyxtq:
Close in favor of #16

PR #16, opened 2019-05-08
Add tox

belldandyxtq:
This PR changes the CI setting to use tox to test different python
versions that are supported by ChainerIO. The latest patched version of
each minor version are tested.
Currently, we test the following:
3.5.7
3.6.8
3.7.2

### DIFF
--- a/.pfnci/test.sh
+++ b/.pfnci/test.sh
@@ -10,4 +10,6 @@ docker run --interactive --rm \
 source /root/.bashrc
 pyenv global 3.5.7 3.6.8 3.7.2
 tox
+pip install sphinx
+cd docs && make html
 EOD
--- a/.pfnci/test.sh
+++ b/.pfnci/test.sh
@@ -5,12 +5,9 @@ gcloud auth configure-docker
 
 docker run --interactive --rm \
        --volume "$(pwd):/repo/" --workdir /repo/ \
-       tianqixu/chainerio \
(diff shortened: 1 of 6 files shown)


belldandyxtq:
pfnci, test this please

.pfnci/test.sh
```diff
@@ -5,12 +5,11 @@ gcloud auth configure-docker
...
-flake8 chainerio
-flake8 tests
-autopep8 -r chainerio tests --diff | tee check_autopep8
-test ! -s check_autopep8
+       chainer/chainerio:latest \
+       bash << EOD
+source /root/.bashrc
+pyenv global 3.5.7 3.6.8 3.7.2
+tox
+pip install sphinx
```
kuenishi: Why don't you make sphinx test and documentation dependencies as new `extras_require` entry in `setup.py` and run as additional deps in tox.ini ?

setup.py
```diff
@@ -30,7 +30,7 @@
     classifiers=[],
     packages=find_packages(),
     package_data={'chainerio' : package_data},
-    extras_require={'test':['pytest', 'flake8']},
+    extras_require={'test':['pytest', 'flake8', 'autopep8']},
```
kuenishi: Not now, but I'd rather feel like moving on to newer and more popular formatters like [black](https://github.com/python/black) .

Issue #9, opened 2019-04-23
Remove workaround after Chainer fix


kuenishi:
https://github.com/chainer/chainer/pull/7036

belldandyxtq:
Solved by #21

PR #18, opened 2019-05-08
Remove workaround

belldandyxtq:
This commit removes the workaround added for the `allow_pickle` issue,
since it has been fixed in chainer in PR #7036.

This solves #9


belldandyxtq:
Close in favour of #21

PR #30, opened 2019-06-25
Add Profiling

belldandyxtq:
With this PR profiling is available to the ChainerIO. Format used in chrome://tracing is also supported, which enables to view the profile in chrome tracing.
The following changes have been made in this PR.
1. Add profile module to ChainerIO to replace the dummy one
2. A decorator has been added to all the I/O functions in ChainerIO for profiling
3. A Chainer like context manager has been added to mange the global and thread local configuration and context
4. A chrome://tracing-compilant format writer to output profile
5. Unit tests for the profile module


kuenishi:
Closing in favor of new V2 API.

PR #32, opened 2019-07-18
Add `eager_open` option to containers

belldandyxtq:
This commit adds `eager_open` option to containers.
When set to True, the container_handler will open the underlying container
in __init__. This behaviour allows users to create a shared underlying objects
before spawning threads, and prevents threads from creating the objects
on their own.

### DIFF
--- a/chainerio/container.py
+++ b/chainerio/container.py
@@ -7,10 +7,12 @@ from typing import Optional
 class Container(IO):
 
     def __init__(self, base_handler: IO, base: str,
-                 io_profiler: Optional[IOProfiler] = None, root: str = ""):
+                 io_profiler: Optional[IOProfiler] = None, root: str = "",
+                 eager_open=False):
         IO.__init__(self, io_profiler, root)
         self.base_handler = base_handler
         self.base = base
+        self.eager_open = eager_open
 
     def reset_base_handler(self, handler: IO) -> None:
         self.base_handler = handler
--- a/chainerio/containers/zip.py
+++ b/chainerio/containers/zip.py
@@ -25,14 +25,17 @@ class ZipFileObject(FileObject):
 
 
 class ZipContainer(Container):
-    def __init__(self, base_handler, base):
-        Container.__init__(self, base_handler, base)
+    def __init__(self, base_handler, base, eager_open=False):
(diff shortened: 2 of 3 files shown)


kuenishi:
Do we have to preserve lazy opening?  My guess is that when the object is created that's the time to open. Then we can remove the `eager_open` option.

chainerio/container.py
```diff
@@ -7,10 +7,12 @@
 class Container(IO):
 
     def __init__(self, base_handler: IO, base: str,
-                 io_profiler: Optional[IOProfiler] = None, root: str = ""):
+                 io_profiler: Optional[IOProfiler] = None, root: str = "",
+                 eager_open=False):
```
kuenishi: Could you explain the possible use case where `eager_open=False` is needed or must be done within the library?  For example, [ZippedImageDataset](https://github.com/chainer/chainer/blob/v6/chainer/datasets/image_dataset.py#L247) already has workaround on issues at MultiprocessIterator and that isn't much complicated. I still don't want to add another keyword argument to container init interface.

belldandyxtq: One example I think we just encountered was: in case of nested zip with Python < 3.7, if we always open zip in `__init__`, i.e. always `eager_open=True` , according to this [PR](https://github.com/chainer/chainerio/pull/38#discussion_r311346659) and our solution in #40 , the whole zip file would be read when creating the handler, as we discussed in that PR, which can be far away from where the first file in zip is actually used. 

I understand your concern about having additional argument in init. This solution can be unnecessary.

kuenishi: Ah I see. Let's finish #40 first and then come back here.

kuenishi:
After #40 merged, I still think `eager_open` should be False by default, or even the option is not needed. Because:
- In future Python 3.7 and later are the majority.
- When the container is opened, that's when the files are needed.
- Lazy opening (e.g. MultiprocessIterator) is not a major use case and can be implemented upon this library.

chainerio/container.py
kuenishi: A nit, but method arguments like this should be in `**kwargs` .

kuenishi:
Now v2 API provides cleaner and general way of lazy opening, ``lazify()`` .

PR #35, opened 2019-07-24
Fix bugs in pfnci script

belldandyxtq:
This commit fixes a bug in pfnci script.
In the previous CI, there are two tests performed, tox for pytest and
sphinx build. The CI script fails to combine all the return value
of the test commands, hence the CI status only show the status of the
sphinx build.

In this commit, the return status of both tests are combined and returned.

### DIFF
--- a/.pfnci/test.sh
+++ b/.pfnci/test.sh
@@ -1,15 +1,14 @@
-#!/usr/bin/env sh
+#!/bin/bash
 set -eux
 
-gcloud auth configure-docker
-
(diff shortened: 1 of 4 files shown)


.pfnci/test.sh
```diff
@@ -1,15 +1,14 @@
-#!/usr/bin/env sh
+#!/bin/bash
```
kuenishi: Wouldn't it be better if `#!/usr/bin/env bash`?

Issue #37, opened 2019-08-06
New design of open wrapper


belldandyxtq:
This is a note for offline discussion:

### Summary:
A new design for open wrapper to replace the current `file_object`

### Background:
In current implementation, a decorator `open_wrapper` wraps the _original file object_ returned by open to a `fileobject`.
During the `__init__`, the `fileobject` further wraps the _original file object_ according to the underlying filesystem and the necessary to add missing functionalities to make them behave like built-in POSIX file objects.
Such `fileobject` is also a base for profiling.

### Problem:
1. Due to lack of `peek` function in `fileobject`, the pickle shows a huge performance degradation. (#36 )
2. Due to lack of `seekable` function, the nested zip container does not work (#34 )
3. Due to lack of `peek` in HDFS, the same problem in #36 would happen when using HDFS.

### Solution:
1. Add all the missing functions.
2. Stop wrapping all the file objects with `fileobject`. Instead, wrap only when it is necessary

### Decision:
We have decided on the 2nd solution.
For the 1st solution, it can be a huge work, and unnecessary for now as the profile module is not ready yet.

### Design:
Extracting the functionality of wrapping the _original file object_ from `__init__` of `fileobject`to a function, and called by the open wrapper. By doing so, we can still wrap the file object when necessary while avoid wrapping everything with `fileobject`.
This new design is supposed to fix #36 and #34.

### Todo List:

- [x]  Replace file object to a wrapper function as described in `Design`. Solves problem 1 and partially for problem 2 (except for Python <= 3.6). (#38)

- [x] Python < 3.6 seekable issue for nested zip as described in #41 . solves Problem 2 for Python <= 3.6 (#40)

- [x] `peek` for HDFS solves solves Problem 3 (#42)

PR #38, opened 2019-08-06
Remove FileObject Wrapper

belldandyxtq:
This PR implements the redesigned open wrapper discussed in #37,
and replace the file object with a `fileobject_returner`
to wrap the file object only when necessary.
For more details, please refer to (Issue #37)

This solves problem 1 and partially for problem 2 (except for Python <= 3.6) in #37


tests/container_tests/test_zip_container.py
```diff
@@ -84,6 +108,7 @@ def test_makedirs(self):
             self.assertRaises(io.UnsupportedOperation,
                               handler.makedirs, "test/test")
 
+    @pytest.mark.timeout(1)
```
kuenishi: Why this timeout is introduced? Introducing non-determinism makes tests very flaky thus must be carefully designed and intended.

chainerio/io.py
```diff
@@ -31,7 +29,26 @@ def __init__(self, io_profiler: Optional[IOProfiler] = None,
         self.io_profiler = io_profiler
         self.type = "BASEIO"
         self.root = root
-        self.fileobj_class = FileObject
+
+    def _fileobject_returner(self, file_obj: Type['IOBase'],
```
kuenishi: `_wrap_fileobject` would be more suitable name?

chainerio/io.py
```diff
@@ -31,7 +29,26 @@ def __init__(self, io_profiler: Optional[IOProfiler] = None,
...
+                             closefd: bool = True,
+                             opener: Optional[Callable[
+                                 [str, int], Any]] = None) -> Type['IOBase']:
+        ''' Replace the file object from the underly system
+
+        This function is called by the open wrapper to check and
+        replace the file object returned by underly system.
+        Derived class overrides this function in order to
+        add functionalities when needed or match the behaviour.
+        In the default case, just forward the given file object.
```
kuenishi: The term 'forward' is not common IMO.

chainerio/containers/zip.py
```diff
@@ -41,15 +29,50 @@ def _check_zip_file_name(self, base):
 
     def _open_zip_file(self, mode='r'):
         mode = mode.replace("b", "")
+        zip_file = self.base_handler.open(self.base, "rb")
+
         if self.zip_file_obj is None:
-            zip_file = self.base_handler.open(self.base, "rb")
+            if isinstance(self.base_handler, ZipContainer) \
+                    and sys.version_info < (3, 7, ):
```
kuenishi: My opinion was not like this; to reopen B.zip and replace the file object right before opening A in B.zip in C.zip, not when opening B.zip in C.zip.

chainerio/filesystems/hdfs.py
```diff
@@ -71,6 +50,21 @@ def _dump_read_file(self, filepath, content):
...
+                             encoding: Optional[str] = None,
+                             errors: Optional[str] = None,
+                             newline: Optional[str] = None,
+                             closefd: bool = True,
+                             opener: Optional[Callable[
+                                 [str, int], Any]] = None) -> Type['IOBase']:
+
+        if 'b' not in mode:
+            file_obj = io.TextIOWrapper(file_obj, encoding, errors, newline)
+
```
kuenishi: Do you work on the issue that PyArrow's file object doesn't have `peek` method here? Or another PR?

chainerio/containers/zip.py
belldandyxtq: Correct me if I am thinking in a wrong way.
>to reopen B.zip and replace the file object right before opening A in B.zip in C.zip, not when opening B.zip in C.zip.
I don't think this is possible because zip file module doesn't do lazy open, hence they check the seekable upon creating the zipfile object. If the file object passed to the zipfile.ZipFile is not seekable, an exception is thrown at that point.

tests/container_tests/test_zip_container.py
belldandyxtq: As I stated in the comment message, I want to make sure the pickle problem is solved in pytest

belldandyxtq: The timeout value need to be chosen again probably

chainerio/filesystems/hdfs.py
belldandyxtq: I would like to do it in another PR since the problem is different

belldandyxtq: The `peek` in HDFS is different from POSIX where the peek was there. For HDFS, we need either add a `peek` method to HDFS file object or create our own pickle.

chainerio/containers/zip.py
belldandyxtq: As we discussed offline, we will remove this issue to another PR

tests/container_tests/test_zip_container.py
belldandyxtq: Move this to another PR

chainerio/containers/zip.py
```diff
@@ -41,15 +29,29 @@ def _check_zip_file_name(self, base):
 
     def _open_zip_file(self, mode='r'):
         mode = mode.replace("b", "")
+        zip_file = self.base_handler.open(self.base, "rb")
+
         if self.zip_file_obj is None:
-            zip_file = self.base_handler.open(self.base, "rb")
```
kuenishi: I don't get why this line is moved above. Could you explain your intention?

chainerio/io.py
```diff
@@ -31,7 +29,26 @@ def __init__(self, io_profiler: Optional[IOProfiler] = None,
...
+                         closefd: bool = True,
+                         opener: Optional[Callable[
+                             [str, int], Any]] = None) -> Type['IOBase']:
+        ''' Replace the file object from the underly system
+
+        This function is called by the open wrapper to check and
+        replace the file object returned by underly system.
+        Derived class overrides this function in order to
+        add functionalities when needed or match the behaviour.
+        In the default case, just return the given file object.
```
kuenishi: maybe "In the default case, it just returns the given file object."

tests/filesystem_tests/test_hdfs_handler.py
```diff
@@ -24,8 +24,6 @@ def test_read_bytes(self):
                 tmpfile.write(self.test_string_b)
             with handler.open(self.tmpfile_name, "rb") as f:
                 self.assertEqual(self.test_string_b, f.read())
-            with handler.open(self.tmpfile_name, "rb") as f:
-                self.assertEqual(self.test_string_b, f.readline())
```
kuenishi: It would be nice if we keep this test, as it can be a test that checks whether the file object is properly wrapped by TextIOWrapper.

belldandyxtq: I think what you meant is tested at https://github.com/chainer/chainerio/blob/09f3194a5419530221ead29da342d023ae03ca1a/tests/filesystem_tests/test_hdfs_handler.py#L35
This test file is opened as "rb".

belldandyxtq:
pfnci, test this please

PR #40, opened 2019-08-14
Nested Zip Support for Python<3.7

belldandyxtq:
This PR adds nested zip support for Python< 3.7.
As described in #41, Unlike the Python > 3.7, due to implementation issues in Python,
the nested zip support is not enabled by default.
This PR introduces a workaround to enable nested zip access in Python <3.7 as follow:

When accessing nested zip with Python< 3.7, the whole nested zip file is read and put into a `BytesIO` object to add `seekable` to the file object to create the corresponding `ZipFile` object. 

This workaround can introduce performance and memory consumption issues when the nested zip file is huge and a small portion of the nested zip is actually needed.
A warning message is generated in that case.

Tests for nested zip are also added in this PR.
This PR solves the 2nd problem in #37, i.e. #41.

### DIFF
--- a/chainerio/containers/zip.py
+++ b/chainerio/containers/zip.py
@@ -47,12 +47,11 @@ class ZipContainer(Container):
                 # for user.
 
                 warnings.warn('In Python < 3.7, '
-                              'To support opeing nested zip as container, '
-                              'Chainerio has to read '
+                              'To support opening nested zip as container, '
+                              'ChainerIO has to read '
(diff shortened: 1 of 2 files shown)


kuenishi:
Please clarify the implementation issue and corresponding workaround at the PR description.

kuenishi:
ping?

belldandyxtq:
Please see #41 as mentioned in the last line of the description.

PR #45, opened 2019-08-16
Fix bugs in `list` in zip

### LINKED ISSUE #43: List all the files in Zip
The current `list` in zip only list the first level of files in zip with the given prefix to match the behavior of POSIX. However, listing all the files in zip is needed in many cases, e.g. read all the files in a zip. We need to have an way to list all the files.

belldandyxtq:
This PR fixes an inconsistency between the API designed behavior
and the actual implementation in `list` of zip.

As described in [__init__.py](https://github.com/chainer/chainerio/blob/65a5d263f72e195bdfdcf33d28944438a902c301/chainerio/__init__.py#L18), in case of containers, in default case when `None` is given as `path_or_prefix`, the `list()` behaves the same as `namelist()` in zipfile.
However, the current implementation does not follow that description and `list()` with `None` return a name list of first level files. Although such behavior matches the POSIX, getting a whole list of files in containers is necessary in many cases as described in #43 .
This PR fixes the such inconsistency and match the behavior to the doc string.

This PR closes #43

### DIFF
--- a/chainerio/containers/zip.py
+++ b/chainerio/containers/zip.py
@@ -79,15 +79,22 @@ class ZipContainer(Container):
         return self._list(path_or_prefix)
 
     def _list(self, path_or_prefix: str = None):
-        _list = set()
-        for name in self.zip_file_obj.namelist():
-            if path_or_prefix and name.startswith(path_or_prefix):
-                name = name[len(path_or_prefix):]
+        if path_or_prefix is None:
+            # in defult case, print the whole list
+            for name in self.zip_file_obj.namelist():
+                yield name
 
-            first_level_file_name = name.split("/")[0]
-            if first_level_file_name and first_level_file_name not in _list:
-                _list.add(first_level_file_name)
-                yield first_level_file_name
+        else:
+            _list = set()
+            for name in self.zip_file_obj.namelist():
+                if path_or_prefix and name.startswith(path_or_prefix):
+                    name = name[len(path_or_prefix):]
+
+                first_level_file_name = name.split("/")[0]
+                if first_level_file_name and \
+                        first_level_file_name not in _list:
+                    _list.add(first_level_file_name)
+                    yield first_level_file_name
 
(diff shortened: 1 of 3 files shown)


kuenishi:
It was past discussion but I think we discussed that we'd introduce `recursive`option to the list method to work with both recursive and non-recursive cases, and the documentation was just temporary workaround, wasn't it? I'd rather think this is not a bug but a work to be done. 

Anyway I agree there's a room to fix this behaviour and the document.

kuenishi:
Also this is not a bug IMO regarding past offline discussion, but just works not yet done.

belldandyxtq:
I think there are two ways to solve this:
1. Add `recursive` parameter to `listdir`
     **pros:**
            - Giving full control to user. Users can choose whether to list `recursively` or not
            - Same behavior in POSIX and zip
    **cons:**
            - Kind of matching POSIX behavior to zip
            - Different interface to [os.listdir](https://docs.python.org/3/library/os.html#os.listdir)
            - The use case of listing the first level in zip is not clear

2. Listing all files in zip with default parameter (this PR)
    **pros:**
            - The same interface to [os.listdir](https://docs.python.org/3/library/os.html#os.listdir)
            - Matching the default behavior of zip(namelist) and (listdir)
   **cons:**
            - Different behavior between `filesystems` and `containers`

What do you think?

kuenishi:
One of the initial motivation of ChainerIO (and thus `chainerio.IO` abstract class) is to provide unified API. Whether container classes' behaviour to `chainerio.IO` is very important question IMO. Aligning its behaviour to Python's `zipfile.ZipFile` isn't that important compared to the value aligning Python's file object behaviour.

Also I think setting `recursive=False` by default we can align its behaviour to `os.listdir()` and it's compatible, rather than saying as different.

belldandyxtq:
I agree, I will add `recursive` in a new PR

Issue #46, opened 2019-08-20
Add recursive to `list`


belldandyxtq:
For containers like zip, it is a common need to list all the files, as discussed in #43. 
To support such operations in a general way, i.e. from API level, we have decided to add `recursive` flag into `list`. When set, `list` returns a generator that will walk though all the files and directories under the given path, similar to os.walk.

To-do

- [x] Add recursive to zip (#47)

- [x] Add recursive to hdfs (#49)

- [x] Add recursive to posix (#57)

PR #48, opened 2019-08-21
Add recursive support to list in http

belldandyxtq:
This PR adds recursive option to http, to align with the API.
This solves the 4th todo in #46


kuenishi:
I would say again that [HTTP](https://tools.ietf.org/html/rfc2616) is just a communication protocol and not a filesystem. By a small chance [WebDAV](https://tools.ietf.org/html/rfc4918) is a filesystem-like protocol built on top of HTTP, but it's almost overtaken by CIFS. That said, the HTTP wrapper should be removed from this library.

belldandyxtq:
close in favor of #51

PR #49, opened 2019-08-21
Add recursive support to list in hdfs

belldandyxtq:
This PR adds recursive support to list in HDFS to align with the API.
This PR solves the 2nd todo in #46

### DIFF
--- a/chainerio/filesystems/hdfs.py
+++ b/chainerio/filesystems/hdfs.py
@@ -108,12 +108,13 @@ class HdfsFileSystem(FileSystem):
             for _dir in dir_list:
                 yield os.path.basename(_dir)
 
-    def _recursive_list(self, prefix, path):
-        for _file in self.connection.ls(path):
-            yield _file[_file.find(prefix):]
+    def _recursive_list(self, path_or_prefix, path):
+        for _file in self.connection.ls(path, detail=True):
+            file_name = _file['name']
+            yield file_name[file_name.find(path_or_prefix):]
 
-            if self.isdir(_file):
-                yield from self._recursive_list(prefix, _file)
+            if 'directory' == _file['kind']:
+                yield from self._recursive_list(path_or_prefix, file_name)
 
     def stat(self, path):
         self._create_connection()
--- a/chainerio/filesystems/hdfs.py
+++ b/chainerio/filesystems/hdfs.py
@@ -95,14 +95,25 @@ class HdfsFileSystem(FileSystem):
         info_str = "using hdfs"
         return info_str
 
-    def list(self, path_or_prefix: str = None):
-        if not path_or_prefix or None is path_or_prefix:
+    def list(self, path_or_prefix: str = None, recursive=False):
+        if path_or_prefix is None:
             path_or_prefix = "/user/{}".format(self.username)
 
         self._create_connection()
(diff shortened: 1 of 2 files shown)


chainerio/filesystems/hdfs.py
```diff
@@ -89,14 +89,25 @@ def info(self):
...
+
+        if recursive:
+            yield from self._recursive_list(path_or_prefix, path_or_prefix)
+        else:
+            dir_list = self.connection.ls(path_or_prefix)
+            for _dir in dir_list:
+                yield os.path.basename(_dir)
+
+    def _recursive_list(self, prefix, path):
+        for _file in self.connection.ls(path):
```
kuenishi: By adding `detail=True` option the `self.isdir()` call below can be removed and the performance would improve a lot.

chainerio/filesystems/hdfs.py
```diff
@@ -89,14 +89,25 @@ def info(self):
...
+            yield _file[_file.find(prefix):]
```
kuenishi: What is this code for?
```
>>> s = 'list'
>>> s[s.find('li'):]
'list'
```

chainerio/filesystems/hdfs.py
belldandyxtq: Thanks, addressed in 4f729ea

chainerio/filesystems/hdfs.py
belldandyxtq: As explained in ad279e8, this is to convert the full URI to relative path from prefix.
"hdfs://nameservice/prefix_dir/testfile" => "prefix_dir/testfile"

chainerio/filesystems/hdfs.py
```diff
@@ -89,14 +94,39 @@ def info(self):
...
+        if path_or_prefix is None:
             path_or_prefix = "/user/{}".format(self.username)
 
         self._create_connection()
-        dir_list = self.connection.ls(path_or_prefix)
-        for _dir in dir_list:
-            yield os.path.basename(_dir)
+        target_dir = self.connection.info(path_or_prefix)
+        if "directory" != target_dir['kind']:
+            return None
```
kuenishi: Why don't you align the behaviour with POSIX filesystem `os.scandir()` ?

chainerio/filesystems/hdfs.py
```diff
@@ -89,14 +94,39 @@ def info(self):
...
+
+        target_path = target_dir['path'] + "/"
+        if not path_or_prefix.endswith("/"):
+            path_or_prefix = path_or_prefix + "/"
```
kuenishi: Stripping trailing `/` at the beginning of this method is a bit smarter than appending itself IMO, as it would be clearer.

chainerio/filesystems/hdfs.py
```diff
@@ -89,14 +94,39 @@ def info(self):
...
+
+        prefix_index = len(target_path[:-len(path_or_prefix)])
```
kuenishi: Both the name prefix and index is confusing here, as it is rather "length" of a prefix, prefix of full path like `hdfs://...` and that is different from `path_or_prefix`. Should be named differently like `print_index`, `prefix_end_index`, or `full_path_prefix_len` .

PR #57, opened 2019-09-17
Add recursive option to POSIX filesystem handler

belldandyxtq:
This PR adds recursive mode to list in posix. This resolve the third todo in #46.

### DIFF
--- a/chainerio/filesystems/posix.py
+++ b/chainerio/filesystems/posix.py
@@ -24,9 +24,18 @@ class PosixFileSystem(FileSystem):
                        buffering, encoding, errors,
                        newline, closefd, opener)
 
-    def list(self, path_or_prefix: str = None):
-        for file in os.scandir(path_or_prefix):
-            yield file.name
+    def list(self, path_or_prefix: str = None, recursive=False):
+        if recursive:
+            path_or_prefix.rstrip("/")
+            # use len instead of len + 1 as root in os.walk does not end
+            # with "/"
+            prefix_end_index = len(path_or_prefix)
+            for root, dirs, files in os.walk(path_or_prefix):
+                for name in files + dirs:
+                    yield os.path.join(root[prefix_end_index:], name)
+        else:
+            for file in os.scandir(path_or_prefix):
+                yield file.name
 
     def stat(self, path):
         return os.stat(path)
--- a/chainerio/filesystems/posix.py
+++ b/chainerio/filesystems/posix.py
@@ -36,13 +36,12 @@ class PosixFileSystem(FileSystem):
                 yield file.name
 
     def _recursive_list(self, prefix_end_index: int, path: str):
-        with os.scandir(path) as dir_it:
-            for entry in dir_it:
-                yield entry.path[prefix_end_index:]
+        for file in os.scandir(path):
+            yield file.path[prefix_end_index:]
 
-                if entry.is_dir():
-                    yield from self._recursive_list(prefix_end_index,
-                                                    entry.path)
+            if file.is_dir():
+                yield from self._recursive_list(prefix_end_index,
+                                                file.path)
 
     def stat(self, path):
         return os.stat(path)
--- a/chainerio/filesystems/posix.py
+++ b/chainerio/filesystems/posix.py
@@ -26,10 +26,9 @@ class PosixFileSystem(FileSystem):
 
     def list(self, path_or_prefix: str = None, recursive=False):
         if recursive:
-            path_or_prefix.rstrip("/")
-            # use len instead of len + 1 as root in os.walk does not end
-            # with "/"
-            prefix_end_index = len(path_or_prefix)
+            path_or_prefix = path_or_prefix.rstrip("/")
+            # plus 1 to include the trailing slash
+            prefix_end_index = len(path_or_prefix) + 1
             yield from self._recursive_list(prefix_end_index, path_or_prefix)
         else:
             for file in os.scandir(path_or_prefix):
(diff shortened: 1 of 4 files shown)


chainerio/filesystems/posix.py
```diff
@@ -24,9 +24,18 @@ def open(self, file_path, mode='r',
...
-    def list(self, path_or_prefix: str = None):
-        for file in os.scandir(path_or_prefix):
-            yield file.name
+    def list(self, path_or_prefix: str = None, recursive=False):
+        if recursive:
+            path_or_prefix.rstrip("/")
+            # use len instead of len + 1 as root in os.walk does not end
+            # with "/"
+            prefix_end_index = len(path_or_prefix)
+            for root, dirs, files in os.walk(path_or_prefix):
```
kuenishi: Single directory listing can be very slow (e.g. overloaded NFS server) and thus I'd prefer recursively calling `os.scandir()` here.

belldandyxtq: I think the `os.walk()` is implemented with `os.scandir()` recursively. 
https://github.com/python/cpython/blob/b9877cd2cc47b6f3512c171814c4f630286279b9/Lib/os.py#L350

kuenishi: I checked the impl, but it's width-first and not depth-first, which would make us feel very slow when number of entries per single directory is huge, while our desired behaviour is depth-first.

belldandyxtq: I am not sure I understand why depth-first is fast. My idea of implementing the depth-first would need a list of directories inside current directory to iterate into. While getting the list of directories needs to call `os.scandir` against current directory, which is the same as width-first.

kuenishi: This  would be a depth-first listing code (and virtually done in hdfs filesystem list):
```python
def rec_list(path):
    for c in os.scandir(path):
        path1 = os.path.join(path, c.name)
        yield path1

        if c.isdir():
            yield from rec_list(path1)
```

kuenishi:
The reason why I prefer depth-first listing is for a case with a directory that includes, say, million children and file system is overloaded by too much meta data, which causes any metadata reference slow. Width-first listing would need waiting for  *all* directory entries fetched from disk across so many inodes, and then listing each directories. While our way of yielding files are depth-first as in both hdfs and zip, and so does `ls(1)`. I believe depth-first printing is more natural and intuitive.

belldandyxtq:
I see your point

chainerio/filesystems/posix.py
```diff
@@ -24,9 +24,24 @@ def open(self, file_path, mode='r',
...
 
-    def list(self, path_or_prefix: str = None):
-        for file in os.scandir(path_or_prefix):
-            yield file.name
+    def list(self, path_or_prefix: str = None, recursive=False):
+        if recursive:
+            path_or_prefix.rstrip("/")
+            # use len instead of len + 1 as root in os.walk does not end
+            # with "/"
+            prefix_end_index = len(path_or_prefix)
```
kuenishi: Needs +1 here.

tests/filesystem_tests/test_posix_handler.py
```diff
@@ -53,8 +53,40 @@ def test_open_non_exist(self):
             self.assertRaises(IOError, handler.open, non_exist_file)
 
     def test_list(self):
+        # directory layout
+        # testlsdir
+        # | - nested_dir1
+        # |   | - nested_dir3
+        # | _ nested_dir2
+        test_dir_name = "testlsdir/"
```
kuenishi: A directory name without trailing slash would have found the bug above...

tests/test_context.py
```diff
@@ -145,6 +145,28 @@ def test_root_fs_override(self):
...
+                                                 nested_dir_name2)
+        chainerio.makedirs(nested_dir_path1)
+        chainerio.makedirs(nested_dir_path2)
+
+        file_list = list(chainerio.list(self.dir_name))
+        self.assertIn(nested_dir_name1, file_list)
+        self.assertIn(self.tmpfile_name, file_list)
+        self.assertNotIn(nested_dir_path2_relative, file_list)
+
+        file_list = list(chainerio.list(self.dir_name, recursive=True))
```
kuenishi: ditto on trailing slash

kuenishi:
Possible performance degrade in depth-first recursion could be stack overflow or cost of keeping directory file descriptors. As I state below maximum order of recursion is limited, both max stack size and max number of file descriptors can be extended, although those numbers won't affect system performance that much.

On max depth of depth-first recursion, these urls ( https://eklitzke.org/path-max-is-tricky , https://stackoverflow.com/q/7140575 ) are very interesting. Practically on Linux and Mac we can assume some rather small max length of the of recursion like 255~4k . On Linux I saw
```
$ find /usr/include/linux -type f| xargs grep PATH_MAX | grep '#define' 
/usr/include/linux/nfs3.h:#define NFS3_MAXPATHLEN               PATH_MAX
/usr/include/linux/un.h:#define UNIX_PATH_MAX   108
/usr/include/linux/btrfs.h:#define BTRFS_INO_LOOKUP_PATH_MAX 4080
/usr/include/linux/btrfs.h:#define BTRFS_INO_LOOKUP_USER_PATH_MAX (4080 - BTRFS_VOL_NAME_MAX - 1)
/usr/include/linux/limits.h:#define PATH_MAX        4096        /* # chars in a path name including nul */
/usr/include/linux/nfs4.h:#define NFS4_MAXPATHLEN               PATH_MAX
/usr/include/linux/netfilter/xt_bpf.h:#define XT_BPF_PATH_MAX           (XT_BPF_MAX_NUM_INSTR * sizeof(struct sock_filter))
/usr/include/linux/netfilter/xt_cgroup.h:#define XT_CGROUP_PATH_MAX     512
```

chainerio/filesystems/posix.py
belldandyxtq: My mistake, sorry. The above line should be `path_or_prefix = path_or_prefix.rstrip("/")`

PR #63, opened 2019-09-25
Remove mutable default argument

kuenishi:
Also I updated some docs of `FileCache`.

### DIFF
--- a/chainerio/cache/file_cache.py
+++ b/chainerio/cache/file_cache.py
@@ -79,7 +79,7 @@ class DummyLock:
 class FileCache(cache.Cache):
     '''Cache system with local filesystem
 
-    Stores cached data in local filesystem,
+    Stores cache data in local temporary files, created in
     ``~/.chainer/chainerio/cache`` by default. Cache data is
     automatically deleted after the object is collected. When this
     object is not correctly closed, (e.g., the process killed by
@@ -87,15 +87,18 @@ class FileCache(cache.Cache):
 
     Arguments:
         length (int): Length of the cache array.
+
         multithread_safe (bool): Defines multithread safety. If this
-            is ``True``, reader-writer locking is introduced behind
-            the cache system. Major use case is with Chainer's
+            is ``True``, reader-writer locking system based on
+            ``threading.Lock`` is introduced behind the cache
+            management. Major use case is with Chainer's
             ``MultithreadIterator``.
+
         do_pickle (bool):
             Do automatic pickle and unpickle inside the cache.
-        dir (str):
-            The path to directory to place cache data in case home
-            directory is not backed by fast storage device.
+
+        dir (str): The path to the directory to place cache data in
+            case home directory is not backed by fast storage device.
 
     TODO(kuenishi): retain cache file in case of correct process
     termination and reuse for future process re-invocation.
--- a/chainerio/cache/file_cache.py
+++ b/chainerio/cache/file_cache.py
@@ -7,7 +7,7 @@ import warnings
 from chainerio import cache
 import pickle
 
-DEFAULT_CACHE_PATH = os.path.join(
+_DEFAULT_CACHE_PATH = os.path.join(
     os.getenv('HOME'), ".chainer", "chainerio", "cache")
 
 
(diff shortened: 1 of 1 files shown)


chainerio/cache/file_cache.py
```diff
@@ -82,16 +82,30 @@ class FileCache(cache.Cache):
     Stores cache data in local filesystem,
     ``~/.chainer/chainerio/cache`` by default. Cache data is
     automatically deleted after the object is collected. If the
-    process quit (typically by SIGTERM) the cache remains after the
+    process quits (typically by SIGTERM) the cache remains after the
```
belldandyxtq: "If the process quits (typically by SIGTERM)*,* the cache remains after the .."

chainerio/cache/file_cache.py
```diff
@@ -82,16 +82,30 @@ class FileCache(cache.Cache):
...
     death of process.
 
+    Arguments:
+        length (int): Length of the cache array. Typically it is
```
belldandyxtq: "Typically, ..."

chainerio/cache/file_cache.py
```diff
@@ -82,16 +82,30 @@ class FileCache(cache.Cache):
...
+            the length of Chainer's ``Dataset``.
+        multithread_safe (bool): Defines multithread safety. If this
+            is ``True``, reader-writer locking is introduced behind
+            the cache system. Typically it is for Chainer's
```
belldandyxtq: "Typically, ..."

belldandyxtq: Should we mention the lock overhead here?

chainerio/cache/file_cache.py
```diff
@@ -82,16 +82,30 @@ class FileCache(cache.Cache):
...
+            ``MultithreadIterator``.
+        do_pickle (bool):
+            Do automatic pickle and unpickle inside the cache.
+        dir (str):
+            The place for cache data in case home directory is not
```
belldandyxtq: The directory where the cache data will be stored. In the default case, the home directory is used. Recommend to fast storage device and avoid using remote filesystems such NFS

chainerio/cache/file_cache.py
```diff
@@ -77,21 +77,33 @@ def unlock(self):
 
 
 class FileCache(cache.Cache):
-    '''Stores cache data in local filesystem
+    '''Cache system with local filesystem
 
-    Stores cache data in local filesystem,
+    Stores cached data in local filesystem,
```
belldandyxtq: "Cache data on local directory, "
As ``~/.chainer/chainerio/cache`` is not a filesystem

chainerio/cache/file_cache.py
```diff
@@ -77,21 +77,33 @@ def unlock(self):
...
-    process quit (typically by SIGTERM) the cache remains after the
-    death of process.
+    automatically deleted after the object is collected. When this
+    object is not correctly closed, (e.g., the process killed by
+    SIGTERM), the cache remains after the death of process.
+
+    Arguments:
+        length (int): Length of the cache array.
+        multithread_safe (bool): Defines multithread safety. If this
+            is ``True``, reader-writer locking is introduced behind
```
belldandyxtq: A readers-writer lock is introduced

chainerio/cache/file_cache.py
```diff
@@ -77,21 +77,33 @@ def unlock(self):
...
+            the cache system. Major use case is with Chainer's
+            ``MultithreadIterator``.
+        do_pickle (bool):
+            Do automatic pickle and unpickle inside the cache.
+        dir (str):
+            The path to directory to place cache data in case home
```
belldandyxtq: https://github.com/chainer/chainerio/pull/63#discussion_r327915198

chainerio/cache/file_cache.py
belldandyxtq: https://github.com/chainer/chainerio/pull/63#discussion_r327914793

kuenishi: Mentioning lock primitive and design is not only for overhead. 1) When designing a multi-threaded system, the locking primitive behind it is very important. For example in Python, `asyncio.Lock` and `threading.Lock` can't be mixed without *practical* level of carefulness. Also it is not easily safe to mix with `multiprocessing.Lock`. Thus I added  description on what primitive is used.
2) The type of lock is important for app developer, too. Understanding that `get` doesn't block another `get` but `put` blocks `get` is important to design multithreaded and read-intensive program like ours. The design would change a lot if it is just a single mutex, where `get` blocks `get`.

chainerio/cache/file_cache.py
kuenishi: I didn't get what you meant. Could elaborate a bit more?

chainerio/cache/file_cache.py
kuenishi: Mentioning lock primitive and design is not only for overhead. 1) When designing a multi-threaded system, the locking primitive behind it is very important. For example in Python, `asyncio.Lock` and `threading.Lock` can't be mixed without *practical* level of carefulness. Also it is not easily safe to mix with `multiprocessing.Lock`. Thus I added  description on what primitive is used.
2) The type of lock is important for app developer, too. Understanding that `get` doesn't block another `get` but `put` blocks `get` is important to design multithreaded and read-intensive program like ours. The design would change a lot if it is just a single mutex, where `get` blocks `get`.

chainerio/cache/file_cache.py
belldandyxtq: Sorry to make it confusing. It was a suggestion, to change the description of "dir" to

```
The directory where the cache data will be stored. In the default case, the home directory is used. Recommend to set the `dir` to a fast storage device and avoid using remote filesystems, such NFS.
```

kuenishi: I think the second sentence is broken, and I think the change is trivial.

PR #65, opened 2019-09-30
Use `klist` to get the username from kerberos

### LINKED ISSUE #50: Use valid username in HDFS
libhdfs hadoop client takes user name from various points, but the priority is from Kerberos ticket. It should not only be taken from operating system like this: https://github.com/chainer/chainerio/blob/master/chainerio/filesystems/hdfs.py#L24

Also a nit: uid is not used and should be removed. https://github.com/chainer/chainerio/blob/master/chainerio/filesystems/hdfs.py#L25

belldandyxtq:
This PR changes to `klist` to get the default principal username from
the default keytab, instead of using the current login username, since
these two sometimes are different.

I have checked Python wrapper of kerberos like 
https://pypi.org/project/python-krbV/
https://pypi.org/project/kerberos/
they are either not maintained or does not support to get the username from keytab

This closes #50

### DIFF
--- a/chainerio/filesystems/hdfs.py
+++ b/chainerio/filesystems/hdfs.py
@@ -18,7 +18,7 @@ logger = logging.getLogger(__name__)
 logger.addHandler(logging.StreamHandler())
 
 
-def parse_klist_output(output):
+def _parse_klist_output(output):
     principle_str = output.decode('utf-8').split('\n')[1]
     klist_principal_pattern = re.compile(
         r'Default principal: (?P<username>.+)@(?P<service>.+)')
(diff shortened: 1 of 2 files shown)


chainerio/filesystems/hdfs.py
```diff
@@ -22,10 +23,24 @@ def __init__(self, io_profiler=None, root="", keytab_path=None):
         self.connection = None
         self.type = 'hdfs'
         self.root = root
-        self.username = getpass.getuser()
-        self.userid = os.getuid()
+        self.username = self._get_principal_name()
         self.keytab_path = keytab_path
         self.nameservice = None
+        print(self.username)
```
kuenishi: Please delete this line

chainerio/filesystems/hdfs.py
```diff
@@ -16,17 +18,51 @@
 logger.addHandler(logging.StreamHandler())
 
 
+def parse_klist_output(output):
```
kuenishi: As this is not to be exposed to users, it would be nice if the name starts with `_`

PR #68, opened 2019-10-02
Bugfix list zip

### LINKED ISSUE #66: Bug in list of directory in zip
This is two bugs when list directory in zip

**ZIP:**
```
tianqi@host:tianqi$ unzip -l test.zip
Archive:  test.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2019-08-16 15:11   test/
       12  2019-07-24 13:13   test/file
        0  2019-08-16 15:11   test/dir/
        0  2019-08-16 15:11   test/dir/nested_dir/
      170  2019-07-24 13:16   test/file.zip
      167  2019-01-16 16:53   python.py
---------                     -------
      349                     6 files
```

**Bug 1**
```
>>> import chainerio
>>> handler = chainerio.open_as_container('test.zip')
>>> list(handler.list('test/'))
['file', 'dir', 'file.zip', 'python.py']
```

Where `python.py` is not in `test/`

**Bug 2**
```
>>> import chainerio
>>> handler = chainerio.open_as_container('test.zip')
>>> list(handler.list('test'))
['python.py']
```

belldandyxtq:
This PR fixes the two bugs mentioned in #66
Closes #66

### DIFF
--- a/tests/container_tests/test_zip_container.py
+++ b/tests/container_tests/test_zip_container.py
@@ -14,10 +14,11 @@ class TestZipHandler(unittest.TestCase):
     def setUp(self):
         # The following zip layout is created for all the tests
         # outside.zip
-        # | - testdir
-        # |   | - nested.zip
-        # |   |   | - nested_dir
-        # |   |   |   | - nested
+        # | - testdir1
+        # |   | - nested1.zip
+        # |       | - nested_dir
+        # |           | - nested
+        # | - testdir2
         # |   | - testfile
         self.test_string = "this is a test string\n"
         self.nested_test_string = \
@@ -25,10 +26,10 @@ class TestZipHandler(unittest.TestCase):
         self.test_string_b = self.test_string.encode("utf-8")
         self.nested_test_string_b = \
             self.nested_test_string.encode("utf-8")
+        self.fs_handler = chainerio.create_handler("posix")
 
         # the most outside zip
         self.zip_file_name = "outside"
-        self.zip_file_path = self.zip_file_name + ".zip"
 
         # nested zip and nested file
         self.tmpdir = tempfile.TemporaryDirectory()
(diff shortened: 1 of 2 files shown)


tests/container_tests/test_zip_container.py
```diff
@@ -97,21 +110,29 @@ def test_list(self):
         with self.fs_handler.open_as_container(self.zip_file_path) as handler:
             zip_generator = handler.list()
             zip_list = list(zip_generator)
-            self.assertIn(self.dir_path.split("/")[0], zip_list)
-            self.assertNotIn(self.zipped_file_name, zip_list)
+            self.assertIn(self.dir_name1.rstrip('/'), zip_list)
+            self.assertIn(self.dir_name2.rstrip('/'), zip_list)
+            self.assertNotIn(self.nested_zipped_file_name, zip_list)
+            self.assertNotIn(self.zipped_file_path, zip_list)
```
kuenishi: One of the causes we missed this bug could be the readability of the test code. I recommend defining expected output as `set()` and compare as a set, like
```python
expected = {self.dir_name1.rstrip('/'), self.dir_name2.rstrip('/')}
zip_list = list(handler.list())
assert expected == set(zip_list)
assert len(expected) == len(zip_list)
```

tests/container_tests/test_zip_container.py
```diff
@@ -97,21 +110,29 @@ def test_list(self):
...
+            # Problem 1, 2 in issue #66
+            dir_list = [self.dir_name1, self.dir_name1.rstrip('/')]
+            for _dir in dir_list:
+                zip_generator = handler.list(_dir)
+                zip_list = list(zip_generator)
+                self.assertNotIn(self.dir_name1.rstrip('/'), zip_list)
+                self.assertNotIn(self.dir_name2.rstrip('/'), zip_list)
+                self.assertNotIn(self.zipped_file_path, zip_list)
+                self.assertIn(self.nested_zip_file_name, zip_list)
+                self.assertNotIn("", zip_list)
```
kuenishi: ditto

tests/container_tests/test_zip_container.py
```diff
@@ -97,21 +110,29 @@ def test_list(self):
...
 
             zip_generator = handler.list(recursive=True)
             zip_list = list(zip_generator)
-            self.assertIn(self.dir_path, zip_list)
-            self.assertIn(os.path.join(self.dir_path, self.zipped_file_name),
-                          zip_list)
+            self.assertIn(self.dir_name1, zip_list)
+            self.assertIn(self.dir_name2, zip_list)
+            self.assertIn(self.zipped_file_path, zip_list)
+            self.assertIn(self.nested_zip_path, zip_list)
```
kuenishi: ditto

tests/container_tests/test_zip_container.py
```diff
@@ -14,21 +14,22 @@ class TestZipHandler(unittest.TestCase):
...
-        # | - testdir
-        # |   | - nested.zip
-        # |   |   | - nested_dir
-        # |   |   |   | - nested
+        # | - testdir1
+        # |   | - nested1.zip
+        # |       | - nested_dir
+        # |           | - nested
+        # | - testdir2
         # |   | - testfile
```
kuenishi: Test case that addresses #66 must be added. Otherwise this commit can't be a test fix.

kuenishi:
Also, this introduces non-compat change on trailing slash, so it would be nice mentioned somewhere at the release time

tests/container_tests/test_zip_container.py
belldandyxtq: The tests are included at https://github.com/chainer/chainerio/blob/11b1b1a50c2abb268585871ea5dfb8ba06921397/tests/container_tests/test_zip_container.py#L119
Although they are not exactly like the examples in #66, But they can reproduce the same errors.
I will refactor the test case to make it more readable

kuenishi:
A few more examples:
```
>>> list(z.list('t'))
['mp']
>>> list(z.list('a'))
['tmp']
>>> list(z.list('/'))
['tmp']
>>> list(z.list('tm'))
['p']
```

belldandyxtq:
For the recursive case, it is known and supposed to be fixed in #72, which has conflicts with this PR and will be rebased after this is merged

belldandyxtq:
I do not think double-slash bug exists in hdfs list as it is not using "prefix matching" as we did here.
```
>>> list(chainerio.list('hdfs:///user/tianqi//tesetdir///'))
['test']
```
It only removes the common prefix of the "service_name" which is a constant across all the realm.

Issue #67, opened 2019-10-01
Accept directory without trailing slash


belldandyxtq:
Related to #61, and #66, we need to accept directory without trailing slash in zip operations
For example:
```
>>> import chainerio
>>> handler = chainerio.open_as_container('test.zip')
>>> handler.isdir('test')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "tianqi/repository/chainerio/chainerio/containers/zip.py", line 130, in isdir
    stat = self.stat(file_path)
  File "tianqi/repository/chainerio/chainerio/containers/zip.py", line 101, in stat
    return self.zip_file_obj.getinfo(path)
  File "tianqi/.pyenv/versions/3.7.2/lib/python3.7/zipfile.py", line 1395, in getinfo
    'There is no item named %r in the archive' % name)
KeyError: "There is no item named 'test' in the archive"
>>> handler.isdir('test/')
True
```

**List of Affected Operations:**

- [x] isdir
- [x] stat
- [x] list
- [x] exists

PR #70, opened 2019-10-02
Support of directories without trailing slash in `stat` of zip

belldandyxtq:
This commit adds non-trailing-slash support to directories to stat of
zip.
The code is mainly from the TODO in #68 
Solves Problem 2 in #67.

### DIFF
--- a/chainerio/containers/zip.py
+++ b/chainerio/containers/zip.py
@@ -98,24 +98,18 @@ class ZipContainer(Container):
 
     def stat(self, path):
         self._open_zip_file()
-        try:
-            stat = self.zip_file_obj.getinfo(path)
-        except KeyError:
+        if path in self.zip_file_obj.namelist():
+            actual_path = path
+        elif (not path.endswith('/')
+              and path + '/' in self.zip_file_obj.namelist()):
             # handles cases when path is a directory but without trailing slash
             # see issue $67
-            if path.endswith('/'):
-                # the case where the file_path already has trailing slash
-                # and it is not in the zip, raise the KeyError
-                raise FileNotFoundError(
-                    "{} is not found".format(path))
-            else:
-                try:
-                    # retry with trailing slash
-                    stat = self.zip_file_obj.getinfo(path + '/')
-                except KeyError:
-                    raise FileNotFoundError(
-                        "{} is not found".format(path))
-        return stat
+            actual_path = path + '/'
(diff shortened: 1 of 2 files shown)


chainerio/containers/zip.py
```diff
@@ -98,7 +98,24 @@ def info(self):
 
     def stat(self, path):
         self._open_zip_file()
-        return self.zip_file_obj.getinfo(path)
+        try:
+            stat = self.zip_file_obj.getinfo(path)
+        except KeyError:
+            # handles cases when path is a directory but without trailing slash
+            # see issue $67
+            if path.endswith('/'):
```
kuenishi: I don't think it's a good practice to write main logic codes in except clause for readability. There should only be error handling. Maybe directly using `NameToInfo` or just turning off the existence flag would be nicer.

chainerio/containers/zip.py
```diff
@@ -98,7 +98,24 @@ def info(self):
 
     def stat(self, path):
         self._open_zip_file()
-        return self.zip_file_obj.getinfo(path)
+        try:
+            stat = self.zip_file_obj.getinfo(path)
```
kuenishi: Maybe appending `/` here if it doesn't finish with `/` would be simpler than handling errors in except clause.

belldandyxtq: We need to check both with `/` (directories) and without `/` (normal files), since we do not know if a path without `/` is a directory or not

belldandyxtq:
I made a todo list to add normpath for the input #76. I will work on it in another PR. In this one I want to focus only on the trailing slash. So do #71,  #69, which have dependency on this.

tests/container_tests/test_zip_container.py
```diff
@@ -254,6 +257,14 @@ def test_nested_zip(self):
                     self.assertEqual(f.read(), self.nested_test_string)
 
     def test_stat(self):
-        # pass for now
-        # TODO(tianqi) add test after we well defined the stat
-        pass
```
kuenishi: Would you mind adding a test to throw `NotADirectoryError` ?

belldandyxtq: It is a test for `stat`, which returns the status of the file or directory, so I do not think the function should throw a `NotADirectoryError`. I will add another test for directory instead.

PR #75, opened 2019-10-04
Add Python 3.5.2 test to CI

### LINKED ISSUE #74: Add Python 3.5.2 tests to CI
As we added support for Python 3.5.2 again, we need to add tests of Python 3.5.2 into CI since the workaround needs to be checked for future PR.

belldandyxtq:
Add Python 3.5.2 test into docker and test script. The Python 3.5.2 replaces the 3.5.8 test as tox does not support specifying micro version. 
Downgrade the base docker image from ubuntu:18.04 to ubuntu:16.04
as 18.04 only support openssl 1.1.0 which does not work with Python
3.5.2.

This closes #74

### DIFF
--- a/.pfnci/docker/Dockerfile
+++ b/.pfnci/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:16.04
 
 LABEL maintainer="tianqi@preferred.jp"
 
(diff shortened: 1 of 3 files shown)


kuenishi:
Why don't you test both 3.5.2 and 3.5.7, which behave differently on type hinting?

belldandyxtq:
>The Python 3.5.2 replaces the 3.5.8 test as tox does not support specifying micro version.

So, tox can specify Python35, 36, 37. If we have different micro versions, only one will be used as Python35. We can do a two round tox, but that would require two tox.ini or run Python36 and 37 twice as well. 
Since the bug is in Python 3.5.2, and the later micro versions are bugfixs, and fixed the bug, testing with Python 3.5.2 might be enough. Also we are testing with Python36 and 37.

PR #88, opened 2019-11-25
Check for the supported scheme in `create_handler`

### LINKED ISSUE #86: Check for supported scheme in `create_handler`
**Problem**
We need to check if the given scheme is supported or not, which is not done currently

**Example**
```
# 'boo' is not a supported scheme
# The call returns h1 as a POSIX (default) handler
# and the root is set to "boo"
>>> h1 = chainerio.create_handler('boo')
>>> h1
<chainerio.filesystems.posix.PosixFileSystem object at 0x2b2848e58eb8>
>>> print(h1.root)
boo
```

belldandyxtq:
This PR add scheme checks to the beginning of `create_handler`, and
raises ValueError when the scheme is not supported.
This PR closes #86.

### DIFF
--- a/chainerio/__init__.py
+++ b/chainerio/__init__.py
@@ -195,6 +195,9 @@ def create_handler(scheme: str) -> IO:
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
+    if not default_context.is_supported_scheme(scheme):
+        raise ValueError("scheme {} is not supported".format(scheme))
+
     (handler, actual_path, is_URI) = \
         default_context.get_handler_by_name(scheme)
     return handler
--- a/chainerio/_context.py
+++ b/chainerio/_context.py
@@ -12,6 +12,9 @@ class FileSystemDriverList(object):
     def __init__(self):
         self._handler_dict = {}
 
+        # TODO(tianqi): dynamically create this list
+        # as well as the patterns upon loading the chainerio module.
+        self.scheme_list = ["hdfs", "posix"]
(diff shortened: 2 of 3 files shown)


kuenishi:
IIRC supported scheme is `hdfs` and `file` as we documented at https://chainerio.readthedocs.io/en/latest/design.html#uri-expression-of-file-paths . The [standard](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) says `file` is registered, while `posix` and `hdfs` are not. I think `hdfs` is general de facto like [Mesos does](https://issues.apache.org/jira/browse/MESOS-4375) but doubt about POSIX. What do you think we should do?

belldandyxtq:
This is a good point. I think we have following three choices:
1. support only "hdfs" and "posix" as it is, despite the fact that "posix" may not match IIRC and our document
2. support "hdfs", "file" and "posix", and point "file" => "posix". But the word "file" can cause confusion.
3. rename the parameter in "create_handler" from 'scheme' to 'fs_type', which matches the intention of the function.

I would support the 3

kuenishi:
IMO 2 looks like a solution to me. We stated at the document to support standard URI-format and adopt the terminology, thus solution 3 will introduce further confusion unless `fs_type` is cleanly and carefully defined. 

But sol.2 is not even perfect and just a workaround. Original confusion that let us find this bug is that a developer gave a path to `create_handler()`. The argument of `create_handler(scheme)` is kind of misleading, as other functions and methods take URI as arguments. That said, more intuitive improvement would be to let it accept full or partial URI and make `.root` member configurable and effective. Also we have to support choosing HDFS-cluster at each handler, which is currently not available.

PR #90, opened 2019-11-27
Add persistence to cache.FileCache

kuenishi:
By letting developers explicitly call `perserve()` method, the cache knows when to freeze the data. This change resolve the issue that cached data can't be reused across the training session (thus re-constructing the cache by re-reading data from remote storage), limiting the use cases to only same number & configuration of cached data.

See `tests/cache_tests/test_file_cache.py` for possible example usage.

### DIFF
--- a/chainerio/cache/file_cache.py
+++ b/chainerio/cache/file_cache.py
@@ -100,9 +100,6 @@ class FileCache(cache.Cache):
         dir (str): The path to the directory to place cache data in
             case home directory is not backed by fast storage device.
 
-    TODO(kuenishi): retain cache file in case of correct process
-    termination and reuse for future process re-invocation.
-
     '''
 
     def __init__(self, length, multithread_safe=False, do_pickle=False,
@@ -126,9 +123,7 @@ class FileCache(cache.Cache):
 
         self.closed = False
         self.indexfp = tempfile.NamedTemporaryFile(delete=True, dir=self.dir)
-        self.indexfile = self.indexfp.name
         self.datafp = tempfile.NamedTemporaryFile(delete=True, dir=self.dir)
-        self.datafile = self.datafp.name
 
         # allocate space to store 2n 64bit unsigned integers
         # 16 bytes * n chunks
@@ -142,15 +137,23 @@ class FileCache(cache.Cache):
             assert r == self.buflen
         self.verbose = verbose
         if self.verbose:
-            print('created index file:', self.indexfile)
-            print('created data file:', self.datafile)
+            print('created index file:', self.indexfp.name)
+            print('created data file:', self.datafp.name)
+
+        self._frozen = False
 
     def __len__(self):
         return self.length
 
     @property
+    def frozen(self):
+        return self._frozen
+
+    @property
     def multiprocess_safe(self):
-        return False
+        # If it's preseved/preloaded, then the file contents are
+        # fixed.
+        return self._frozen
 
     @property
     def multithread_safe(self):
@@ -178,6 +181,7 @@ class FileCache(cache.Cache):
             return data
 
     def put(self, i, data):
+        assert not self._frozen
         try:
             if self.do_pickle:
                 data = pickle.dumps(data)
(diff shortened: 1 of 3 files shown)


belldandyxtq:
I think this closes #3

chainerio/cache/file_cache.py
```diff
@@ -251,3 +255,56 @@ def close(self):
                 self.datafp.close()
                 self.indexfp = None
                 self.datafp = None
+
+    def preload(self, name):
```
belldandyxtq: Is it possible to split the `preload` and the `freeze` of cache?
In the current implementation, user has to do following to `freeze` the cache and get multi-process safe:
load => preserve => preload

If we can split these two functionalities, then we can simply do "load" => "freeze". Moreover, the cache can be modified before "freeze".

chainerio/cache/file_cache.py
```diff
@@ -251,3 +255,56 @@ def close(self):
...
+            # Hard link and save them
+            self.indexfp.close()
+            self.datafp.close()
+
+            self.indexfp = open(indexfile, 'rb')
+            self.datafp = open(datafile, 'rb')
+            self._frozen = True
+
+    def preserve(self, name):
+        '''Preserve the cache as persistent files in the disk
```
belldandyxtq: on the disk

chainerio/cache/file_cache.py
```diff
@@ -251,3 +255,56 @@ def close(self):
...
+
+        Once the cache is preserved, cache files are not to be removed
```
belldandyxtq: cache files are not to be removed => cache files will not be removed

chainerio/cache/file_cache.py
kuenishi: What do you mean by load?
If "load" means writing data into cache files, that means creating file cache with specific *name* if I understand it correctly. But it is unacceptable because it leads to partially written *named* files left after job failure and they'll definitely mess up the cache and following re-run of the job.

Also the principle of cache in this module is data being immutable and deterministic, there should be no modification. What do you mean by "modified"?

belldandyxtq: Sorry to cause confusion, by "load" I mean reading data from remote storage. If I understand correctly, in order to use this module, user has to do following:

1. read data from remote storage
2. preserve
3. preload

My suggestion is to have a "freeze" function, so that we can:

1. read data from remote storage
2. freeze

to support the multiprocess. I do not think it would increase the risk of having a broken cache.

By modifying, I do not mean changing the data itself, but the ability of adding new cache entities.

kuenishi: > I do not think it would increase the risk of having a broken cache.
I believe definitely it happens. For example,

1. Create or a cache file named "foo.cache"
2. Write 50% data to "foo.cache", like `put(1, b'bar')`
3. Job fail

Even in this case partial write may happen like `b'ba'` was written to disk but `b'r'` wasn't. Then, restart the job,

1. Open a 50% written cache file named "foo.cache"
2. Can't run `put(1, b'bar')` as the data 1 is immutable
3. Job fail

Also in this case it is not possible to put data nor to run `get(1)` if partial write happened. This is typical scinario of broken data and broken cache files must be thrown away.

belldandyxtq: I think we both are misunderstanding each other. I think in your code, you integrated the "freeze" into preserve, which I misunderstood, so the 3 is not necessary. But I think that design forces the user to preserve the cache, which is not always necessary. For example, they just want to enable multiprocessing, which only needs "freezing" the cache.

In my example, the cache should not be frozen until all the data is loaded, and after "freeze", no data can be added to the cache. Then when the job fails at a "put", which is definitely before the "freeze", the cache will be neither frozen nor preserved. That is why I think such design will not increase the risk of having broken case. 

Sorry for making confusion.

kuenishi: Automatic deletion of the cache among multiple processes does not work correctly because there are no synchronization on closing the file objects, where child processes can't read contents after parent process exit. This is because Python's `tempfile` module explicitly unlinks named temporary files. Please try following script:
```python
import tempfile
import os
import multiprocessing as mp
import time

with tempfile.NamedTemporaryFile() as f:
    f.write(b'foo')
    f.flush()
    name = f.name

    pid = os.fork()
    print('pid', pid)
    if 0 == pid:
        pname = 'child'
    else:
        pname = 'parent'

    f2 = open(name, 'rb')
    assert f2
    time.sleep(1)
    if pname == 'child':
        time.sleep(1)
```
You'll see this:
```
$ Traceback (most recent call last):
  File "t.py", line 22, in <module>
    time.sleep(1)
  File "/usr/lib/python3.7/tempfile.py", line 500, in __exit__
    self.close()
  File "/usr/lib/python3.7/tempfile.py", line 507, in close
    self._closer.close()
  File "/usr/lib/python3.7/tempfile.py", line 444, in close
    unlink(self.name)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmplpb16mgk'
```

belldandyxtq: I see, thank you for the explanation. Can you fix the other issues?

PR #91, opened 2019-11-29
Use keytab by environment variable

kuenishi:


chainerio/filesystems/hdfs.py
```diff
@@ -40,7 +40,11 @@ def __init__(self, io_profiler=None, root="", keytab_path=None):
         if self.username is None:
             # in case klist fails, use the login username instead
             self.username = self._get_login_username()
+
         self.keytab_path = keytab_path
+        if self.keytab_path is None:
+            self.keytab_path = os.getenv("CHAINERIO_KEYTAB")
```
belldandyxtq: How about renaming it "CHAINERIO_HDFS_KEYTAB"?

docs/source/reference.rst
```diff
@@ -28,14 +28,22 @@ Toplevel Functions
 .. autoclass:: chainerio.container.Container
    :members:
 
+
+.. note:: With environment variable
+          ``CHAINERIO_KEYTAB=path/to/your.keytab`` set, ``hdfs``
```
belldandyxtq: ditto

kuenishi:
I have changed the whole design to use `KRB5_KTNAME` , env variable for `kinit` which is supported by krbticket 1.0.1 .

chainerio/filesystems/hdfs.py
```diff
@@ -67,11 +68,12 @@ def _create_connection(self):
...
-            if None is not self.keytab_path:
-                self.ticket = KrbTicket.init(
-                    self.username,
-                    self.keytab_path)
-                self.ticket.updater_start()
+            # Updater automatically let kinit take ``KRB5_KTNAME``
+            # variable. If /etc/krb5.keytab doesn't exist, krbticket
+            # tries to update the ticket with ``kinit -R`` as much as
+            # possible.
+            self.ticket = KrbTicket.init(self.username)
```
belldandyxtq: Is the update needed when we do not have the keytab?

kuenishi: Please read the code comment above

belldandyxtq: I meant do we have to update the ticket even if we do not have the keytab?

kuenishi: Yes. Ticket expiration date can be renewable until renewable date by `kinit -R` and the former is usually earlier than latter, like hours or days.

belldandyxtq: I see

PR #97, opened 2019-12-11
Bugfix: get correct principal name when keytab is given

### LINKED ISSUE #93: Bugs in getting HDFS Principal Name when keytab is given
**Problem statement**
Getting default Principal name fails when "KRB5_KTNAME" is given

**How to reproduce**
```
>>> import os
>>> import chainerio
>>> os.getenv("KRB5_KTNAME")
'/etc/keytab/user.keytab'
>>> chainerio.exists("hdfs:///user/tianqi")
kinit: Keytab contains no suitable keys for user_xxx@NAMESERVICE while getting initial credentials
```

**Cause**
https://github.com/chainer/chainerio/blob/9ef197461fd715f1fe90b10bf0090b9a0c1849c2/chainerio/filesystems/hdfs.py#L52
`klist -c` takes cache instead of keytab. 

We should also create an interface for user to specify their username instead of always using the default.

belldandyxtq:
This commit fixes a bug where the principal name is not obtained
correctly when a keytab is given though KRB5_KTNAME

This closes #93 
This should be tested after #96

### DIFF
--- a/chainerio/filesystems/hdfs.py
+++ b/chainerio/filesystems/hdfs.py
@@ -51,11 +51,7 @@ def _parse_principal_name_from_keytab(output):
 
 
 def _get_principal_name_from_keytab():
-    keytab_path = os.getenv("KRB5_KTNAME")
-    if keytab_path is None:
-        return None
-
-    output = _run_klist(keytab_path)
+    output = _run_klist(use_keytab=True)
     if output is None:
         return None
 
@@ -69,11 +65,11 @@ def _get_principal_name_from_klist():
     return _parse_principal_name_from_klist(output.decode('utf-8'))
 
 
-def _run_klist(keytab_path=None):
+def _run_klist(use_keytab=False):
     try:
         command = ['klist']
-        if keytab_path is not None:
-            command += ['-k', keytab_path]
+        if use_keytab:
+            command += ['-k']
         pipe = subprocess.Popen(command, stdout=subprocess.PIPE,
                                 stderr=subprocess.PIPE)
         out, err = pipe.communicate()
@@ -106,9 +102,9 @@ class HdfsFileSystem(FileSystem):
             principal_name = _get_principal_name_from_keytab()
             if principal_name is not None:
                 return principal_name
-            else:
-                # in case klist fails, use the login username instead
-                return self._get_login_username()
+
+        # in case every thing, use the login username instead
+        return self._get_login_username()
 
     def _get_login_username(self):
         return getpass.getuser()
--- a/chainerio/filesystems/hdfs.py
+++ b/chainerio/filesystems/hdfs.py
@@ -93,15 +93,16 @@ class HdfsFileSystem(FileSystem):
         self.nameservice = None
 
     def _get_principal_name(self):
-        # get the default principle name from `klist` cache
+        # get the default principal name from `klist` cache
         principal_name = _get_principal_name_from_klist()
 
         if principal_name is not None:
             return principal_name
-        else:
-            principal_name = _get_principal_name_from_keytab()
-            if principal_name is not None:
-                return principal_name
+
+        # try getting principal name from keytab
+        principal_name = _get_principal_name_from_keytab()
+        if principal_name is not None:
+            return principal_name
 
         # in case every thing, use the login username instead
         return self._get_login_username()
--- a/chainerio/filesystems/hdfs.py
+++ b/chainerio/filesystems/hdfs.py
@@ -18,8 +18,12 @@ logger = logging.getLogger(__name__)
 logger.addHandler(logging.StreamHandler())
 
 
-def _parse_klist_output(output):
-    principle_str = output.decode('utf-8').split('\n')[1]
+def _parse_principal_name_from_klist(output):
+    output_array = output.split('\n')
+    if len(output_array) < 2:
+        return None
+
+    principle_str = output_array[1]
     klist_principal_pattern = re.compile(
         r'Default principal: (?P<username>.+)@(?P<service>.+)')
     ret = klist_principal_pattern.match(principle_str)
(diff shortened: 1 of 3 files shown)


tests/filesystem_tests/test_hdfs_handler.py
```diff
@@ -53,6 +54,40 @@ def test_klist_not_exist(self):
...
+        with chainerio.create_handler(self.fs) as handler:
+            self.assertEqual(dummy_username, handler.username)
+
+        # put KRB5_KTNAME back
+        if original_krb5_ktname is None:
+            del os.environ['KRB5_KTNAME']
+        else:
+            os.environ['KRB5_KTNAME'] = original_krb5_ktname
+
+        chainerio.remove(keytab_path)
```
kuenishi: Use `tempfile.TemporaryDirectory()` with it's cleanup capability on `with` statement in case of unexpected test failure.

tests/filesystem_tests/test_hdfs_handler.py
```diff
@@ -53,6 +54,40 @@ def test_klist_not_exist(self):
...
+        out, err = pipe.communicate()
+        return keytab_path
+
+    def test_get_principel_name_from_keytab(self):
+        dummy_username = "IAmADummy"
+        keytab_path = self.create_dummy_keytab(dummy_username)
+
+        # save the original KRB5_KTNAME
+        if "KRB5_KTNAME" in os.environ:
+            original_krb5_ktname = os.environ['KRB5_KTNAME']
```
kuenishi: Use `os.getenv()` or `os.environ.get()` for default value.

tests/filesystem_tests/test_hdfs_handler.py
```diff
@@ -53,6 +54,40 @@ def test_klist_not_exist(self):
...
+        else:
+            original_krb5_ktname = None
+
+        os.environ['KRB5_KTNAME'] = keytab_path
+        with chainerio.create_handler(self.fs) as handler:
+            self.assertEqual(dummy_username, handler.username)
+
+        # put KRB5_KTNAME back
+        if original_krb5_ktname is None:
+            del os.environ['KRB5_KTNAME']
```
kuenishi: Also this must be done in `finally` clause in case of unexpected exceptions.

tests/filesystem_tests/test_hdfs_handler.py
```diff
@@ -53,6 +54,40 @@ def test_klist_not_exist(self):
...
+        command = "(echo 'addent -password -p {}@{} -k 1 -e rc4-hmac' &&\
+                  sleep 1 && echo {} && echo write_kt {}) \
+                  | ktutil".format(dummy_username, "dummy_realm",
+                                   dummy_password, keytab_path)
+        pipe = subprocess.Popen(command, stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE, shell=True)
+        out, err = pipe.communicate()
+        return keytab_path
+
+    def test_get_principel_name_from_keytab(self):
```
kuenishi: This method name indicates it being intended for unittest of `def _parse_klist_keytab_output(output):` somehow. Also practical unittest on parsing the output of `klist` had better be separated from integrated test of `create_handler()`.

belldandyxtq:
The username is not only taken from keytab. If the keytab is given then we get username from keytab, otherwise from `klist`. If both failed, from login name.

kuenishi:
> The username is not only taken from keytab. If the keytab is given then we get username from keytab, otherwise from `klist`. If both failed, from login name.

Yes. But the priority behaviour is different than mine. Current behaviour is 1) if KRB5_KTNAME is defined then run `kinit -k` and take name from keytab. 2) otherwise just run `klist` implicitly taking `KRB5CCNAME` and take name from credential file. This is different from KrbTicket behaviour and I'm worrying about potential bug that hides behind this gap. Also, with my suggestion the code would be much simpler:
1. try taking name form `klist`
2. if it failed with exception then run `klist -k`
3. if it failed with exception then take the name from Unix username.

tests/filesystem_tests/test_hdfs_handler.py
```diff
@@ -53,6 +56,36 @@ def test_klist_not_exist(self):
...
+        dummy_username = "IAmADummy"
+        with tempfile.TemporaryDirectory() as tmpd:
+            # save the original KRB5_KTNAME
+            original_krb5_ktname = os.environ.get('KRB5_KTNAME')
+            try:
+                keytab_path = self.create_dummy_keytab(tmpd, dummy_username)
+
+                os.environ['KRB5_KTNAME'] = keytab_path
+                with HdfsFileSystem() as handler:
```
kuenishi: My intention was not to change from `create_handler` to direct object construction, but to clarify what unit feature to be tested with the test method name matching test target. Hereby there're several unit functionalities here, 1) calling the klist command, 2) parsing klist command output, 3) identifying the user name. Strictly speaking, all these three must be tested one by one in unit tests. I don't request strict tests, but at least it must be clarified what is being tested here.

In other word, `_parse_klist_output()` is tested at (new) line 93 below, but why not `_parse_klist_keytab_output(out)` ? Also integrated test of those would be nice if it's done here.

chainerio/filesystems/hdfs.py
```diff
@@ -30,36 +34,81 @@ def _parse_klist_output(output):
...
+
+
+def _get_principal_name_from_klist():
+    output = _run_klist()
+    if output is None:
+        return None
+    return _parse_principal_name_from_klist(output.decode('utf-8'))
+
+
+def _run_klist(keytab_path=None):
```
kuenishi: I think passing boolean indicating `-k` rather than string would remove potential bug that may stem from string variations.

chainerio/filesystems/hdfs.py
```diff
@@ -30,36 +34,81 @@ def _parse_klist_output(output):
...
+    ret = keytab_principle_pattern.match(principle_str)
+    if ret:
+        pattern_dict = ret.groupdict()
+        return pattern_dict['username']
+    else:
+        return None
+
+
+def _get_principal_name_from_keytab():
+    keytab_path = os.getenv("KRB5_KTNAME")
```
kuenishi: I don't this line is needed, but just running `klist -k` is enough. Because not only it automatically refers to `KRB5_KTNAME`but follows default fallback to `/etc/krb5.keytab` .

chainerio/filesystems/hdfs.py
```diff
@@ -30,36 +34,81 @@ def _parse_klist_output(output):
...
-                                    stderr=subprocess.PIPE)
-            out, err = pipe.communicate()
-            if out == b'' and err != b'':
-                return None
+        # get the default principle name from `klist` cache
+        principal_name = _get_principal_name_from_klist()
+
+        if principal_name is not None:
+            return principal_name
+        else:
```
kuenishi: This else indent is not needed, I would prefer it removed for readability.

kuenishi: ping?

PR #99, opened 2019-12-12
Bugfix: Reflect default handler in global functions

### LINKED ISSUE #98: Some global functions do not follow the `set_root`
**Problems**
A few `chainer.xxx` functions (e.g. `chainerio.exists`) do not  follow the `set_root`.

**How to reproduce**
```
 hdfs dfs -ls /home/
ls: `/home/': No such file or directory
$ python
Python 3.6.8 (default, Oct  7 2019, 09:28:08)                
[GCC 7.4.0] on linux                                                                                                                                          
Type "help", "copyright", "credits" or "license" for more information.      
>>> import chainerio
>>> chainerio.set_root("hdfs")
>>> chainerio.exists("/home/")
True
```
**Cause**
These functions are calling an old function which ignores the `default_handler`
https://github.com/chainer/chainerio/blob/ca79de952f21cc216cc70c1a0b87afd6d07426cd/chainerio/__init__.py#L267
https://github.com/chainer/chainerio/blob/ca79de952f21cc216cc70c1a0b87afd6d07426cd/chainerio/__init__.py#L43

belldandyxtq:
This PR fixes a bug that some of the global functions do not refer
to the default handler.
This PR closes #98

### DIFF
--- a/chainerio/__init__.py
+++ b/chainerio/__init__.py
@@ -41,7 +41,7 @@ def list(path_or_prefix: Optional[str] = None,
     if None is path_or_prefix:
         path_or_prefix = ""
 
-    (handler, actual_path) = default_context.get_handler(path_or_prefix)
+    (handler, actual_path) = default_context.get_or_create_default_handler(path_or_prefix)
     return handler.list(actual_path, recursive)
 
 
@@ -57,7 +57,7 @@ def info() -> str:
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, dummy_path) = default_context.get_handler()
+    (handler, dummy_path) = default_context.get_or_create_default_handler()
     return handler.info()
 
 
@@ -81,7 +81,7 @@ def open(file_path: str, mode: str = 'rb',
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_handler(file_path)
+    (handler, actual_path) = default_context.get_or_create_default_handler(file_path)
     return handler.open(
         actual_path, mode, buffering, encoding,
         errors, newline, closefd, opener)
@@ -212,7 +212,7 @@ def isdir(path: str) -> bool:
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_handler(path)
+    (handler, actual_path) = default_context.get_or_create_default_handler(path)
     return handler.isdir(actual_path)
 
 
@@ -228,7 +228,7 @@ def mkdir(path: str, mode: int = 0o777, *,
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_handler(path)
+    (handler, actual_path) = default_context.get_or_create_default_handler(path)
     return handler.mkdir(actual_path, mode, dir_fd=dir_fd)
 
 
@@ -244,7 +244,7 @@ def makedirs(path: str, mode: int = 0o777,
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_handler(path)
+    (handler, actual_path) = default_context.get_or_create_default_handler(path)
     return handler.makedirs(actual_path, mode, exist_ok)
 
 
@@ -259,7 +259,7 @@ def exists(path: str) -> bool:
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_handler(path)
+    (handler, actual_path) = default_context.get_or_create_default_handler(path)
     return handler.exists(actual_path)
 
 
@@ -275,8 +275,8 @@ def rename(src: str, dst: str) -> None:
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler_src, actual_path_src) = default_context.get_handler(src)
-    (handler_dst, actual_path_dst) = default_context.get_handler(dst)
+    (handler_src, actual_path_src) = default_context.get_or_create_default_handler(src)
+    (handler_dst, actual_path_dst) = default_context.get_or_create_default_handler(dst)
     # TODO: containers are not supported here
     if type(handler_src) != type(handler_dst):
         raise NotImplementedError(
@@ -302,7 +302,7 @@ def remove(path: str, recursive: bool = False) -> None:
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_handler(path)
+    (handler, actual_path) = default_context.get_or_create_default_handler(path)
     return handler.remove(actual_path, recursive)
 
 
--- a/chainerio/__init__.py
+++ b/chainerio/__init__.py
@@ -41,7 +41,8 @@ def list(path_or_prefix: Optional[str] = None,
     if None is path_or_prefix:
         path_or_prefix = ""
 
-    (handler, actual_path) = default_context.get_or_create_default_handler(path_or_prefix)
+    (handler, actual_path) = \
+        default_context.get_or_create_default_handler(path_or_prefix)
     return handler.list(actual_path, recursive)
 
 
@@ -81,7 +82,8 @@ def open(file_path: str, mode: str = 'rb',
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_or_create_default_handler(file_path)
+    (handler, actual_path) = \
+        default_context.get_or_create_default_handler(file_path)
     return handler.open(
         actual_path, mode, buffering, encoding,
         errors, newline, closefd, opener)
@@ -212,7 +214,8 @@ def isdir(path: str) -> bool:
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_or_create_default_handler(path)
+    (handler, actual_path) = \
+        default_context.get_or_create_default_handler(path)
     return handler.isdir(actual_path)
 
 
@@ -228,7 +231,8 @@ def mkdir(path: str, mode: int = 0o777, *,
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_or_create_default_handler(path)
+    (handler, actual_path) = \
+        default_context.get_or_create_default_handler(path)
     return handler.mkdir(actual_path, mode, dir_fd=dir_fd)
 
 
@@ -244,7 +248,8 @@ def makedirs(path: str, mode: int = 0o777,
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path) = default_context.get_or_create_default_handler(path)
+    (handler, actual_path) = \
+        default_context.get_or_create_default_handler(path)
     return handler.makedirs(actual_path, mode, exist_ok)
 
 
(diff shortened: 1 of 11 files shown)


kuenishi:
Please fix the tests.

belldandyxtq:
The tests are fixed

chainerio/__init__.py
```diff
@@ -213,8 +213,7 @@ def isdir(path: str) -> bool:
     global _DEFAULT_CONTEXT
     default_context = _DEFAULT_CONTEXT
 
-    (handler, actual_path, is_URI) = \
-        default_context.get_handler_by_name(path)
+    (handler, actual_path) = default_context.get_handler(path)
```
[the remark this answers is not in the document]
belldandyxtq: I have implemented a cache for handlers used by the global context, so the handlers will be cached and do not need to closed every time

chainerio/_context.py
```diff
@@ -62,55 +57,52 @@ def is_supported_scheme(self, scheme: str) -> bool:
 class DefaultContext(object):
     def __init__(self):
         self._fs_handler_list = FileSystemDriverList()
-        self._root = ""
 
         self._default_context = \
-            self._fs_handler_list.get_handler_for_root("posix")[0]
+            self._fs_handler_list.get_handler("posix")[0]
+
+        self._global_handler_cache = {}
```
kuenishi: We know `DefaultContext` is global object so the name of this object could be just `_handler_cache` or simply `_handlers` . Even if any of these names adopted, what is stored in this dict is not clear so it must be documented in the comment ("global handler cache" isn't even enough from the viewpoint of readability) .

chainerio/_context.py
```diff
@@ -18,40 +18,35 @@ def __init__(self):
...
+        else:
+            actual_path = path
+            is_URI = False
 
         handler = create_fs_handler(fs_type)
         return (handler, actual_path, is_URI)
 
-    def get_handler_for_root(self,
-                             uri_or_handler_name: str) -> Tuple[IO, str, bool]:
+    def get_handler(self, uri_or_handler_name: str) -> Tuple[IO, str, bool]:
```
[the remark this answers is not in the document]
belldandyxtq: removed in 5502f8b

chainerio/_context.py
belldandyxtq: resolved in 228f0ef

chainerio/_context.py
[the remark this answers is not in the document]
belldandyxtq: resolved in de9c744

chainerio/_context.py
belldandyxtq: resolved in ba72759

belldandyxtq:
I am not sure what you mean by thread safe on `_global_handler_cache`. Since the operation on this dictionary is just assignment and lookup, I think it is thread safe via GIL.

tests/test_context.py
```diff
@@ -129,19 +131,28 @@ def test_root_local_override(self):
             with chainerio.open('file://' + __file__) as fp:
                 self.assertEqual(fp.read(), my_script.read().encode("utf-8"))
 
+        chainerio.set_root('posix')
```
kuenishi: Ditto

tests/test_context.py
```diff
@@ -153,9 +164,24 @@ def test_root_fs_override(self):
...
+
+        chainerio.rename(hdfs_tmpfile, hdfs_tmpfile_renamed)
+        self.assertEqual(chainerio.exists(hdfs_tmpfile), False)
+        self.assertEqual(chainerio.exists(hdfs_tmpfile_renamed), True)
+
+        chainerio.remove(hdfs_tmpdir, recursive=True)
+        self.assertEqual(chainerio.exists(hdfs_tmpdir), False)
         conn.close()
 
+        chainerio.set_root("posix")
```
kuenishi: Ditto

tests/test_context.py
```diff
@@ -153,9 +164,24 @@ def test_root_fs_override(self):
...
+        self.assertEqual(chainerio.exists(hdfs_tmpfile), True)
+        self.assertEqual(chainerio.exists(hdfs_tmpfile_renamed), False)
+
+        chainerio.rename(hdfs_tmpfile, hdfs_tmpfile_renamed)
+        self.assertEqual(chainerio.exists(hdfs_tmpfile), False)
+        self.assertEqual(chainerio.exists(hdfs_tmpfile_renamed), True)
+
+        chainerio.remove(hdfs_tmpdir, recursive=True)
+        self.assertEqual(chainerio.exists(hdfs_tmpdir), False)
         conn.close()
```
kuenishi: This kind of resource collection must be cleanly done with `try: .. finally: ...`

tests/test_context.py
```diff
@@ -129,19 +131,28 @@ def test_root_local_override(self):
...
 
+        chainerio.set_root('posix')
+
     # override with different filesystem
     @unittest.skipIf(shutil.which('hdfs') is None, "HDFS client not installed")
     def test_root_fs_override(self):
         from pyarrow import hdfs
 
-        hdfs_tmpfile = "tmpfile_hdfs"
+        hdfs_tmpdir = "tmpdir_hdfs"
```
kuenishi: I don't think we need all these `hdfs_` prefix and `_hdfs` suffix, as these tests are all about HDFS.

belldandyxtq: resolved in 7423eaa

tests/test_context.py
belldandyxtq: resolved in a8fa100

chainerio/_context.py
```diff
@@ -14,46 +13,35 @@ def __init__(self):
...
         else:
-            (new_handler, actual_path, is_URI) = self.get_handler_from_path(
-                uri_or_handler_name)
-            new_handler.root = actual_path
-            return (new_handler, actual_path, is_URI)
+            # create a new handler
+            handler = create_fs_handler(fs_type)
+            self._handler_cache[fs_type] = handler
+
+        return handler
```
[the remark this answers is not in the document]
belldandyxtq: resolved in bca2055

chainerio/_context.py
```diff
@@ -62,55 +50,40 @@ def is_supported_scheme(self, scheme: str) -> bool:
 class DefaultContext(object):
     def __init__(self):
         self._fs_handler_list = FileSystemDriverList()
```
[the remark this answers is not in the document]
belldandyxtq: resolved in 5f6164f

chainerio/_context.py
```diff
@@ -14,46 +13,35 @@ def __init__(self):
...
-            return (path, False)
+                return (fs_type, ret.groupdict()["path"], True)
 
-    def get_handler_from_path(self, path: str) -> Tuple[IO, str, bool]:
-        (fs_type, actual_path, is_URI) = self._determine_fs_type(path)
+        return ("posix", path, False)
 
-        handler = create_fs_handler(fs_type)
-        return (handler, actual_path, is_URI)
+    def get_handler(self, fs_type: str) -> Tuple[IO]:
```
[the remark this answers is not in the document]
belldandyxtq: resolved in a64a2e4.

There were two `get_handler`s, one was in the `DefaultContext` which returns `Tuple[IO, str]` , the other was in the `FileSystemHandlerList`, which returns `IO`. I renamed them to `get_or_create_cached_handler` and `get_or_create_default_handler` respectively.

tests/test_context.py
```diff
@@ -32,6 +32,8 @@ def test_set_root(self):
         with chainerio.open(self.tmpfile_name, "r") as fp:
             self.assertEqual(fp.read(), self.test_string_str)
 
+        chainerio.set_root('posix')
```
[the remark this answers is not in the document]
belldandyxtq: resolved in a8fa1007ba8335bd80910b7993ded7ef66f1de74

tests/filesystem_tests/test_hdfs_handler.py
```diff
@@ -37,6 +37,58 @@ def setUp(self):
...
+                # mkdirs_name = "mkdirs/dirs"
+                # handler.makedirs(mkdirs_name)
+                # self.assertTrue(handler.exists(mkdirs_name))
+                # isdir with root
+                self.assertTrue(handler.isdir(mkdir_name))
+                # remove with root
+                handler.remove(mkdir_name)
+                self.assertFalse(handler.exists(mkdir_name))
+            finally:
+                handler.root = ""
```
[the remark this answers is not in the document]
belldandyxtq: The `.root` is set back to delete the `testdir` as since the `.root` was set to `testdir` at https://github.com/chainer/chainerio/blob/dea9e5a727d87caf94d7341264bd7cec41e193dd/tests/filesystem_tests/test_hdfs_handler.py#L57

Hence we need to move to the default root before deleting the `testdir`. I added a comment to improve the readability

tests/test_context.py
belldandyxtq: resolved in a8fa100

tests/test_context.py
belldandyxtq: resolved in a8fa100

belldandyxtq:
I agree that we can introduce mypy test. However, only part of the code have the type annotation, we need first add type annotation first. What do you think?

chainerio/_context.py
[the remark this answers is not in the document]
kuenishi: Please use context manager with `with` clause to make sure lock is released.

chainerio/filesystems/hdfs.py
```diff
@@ -84,14 +84,27 @@ def _run_klist(use_keytab=False):
 
 class HdfsFileSystem(FileSystem):
     def __init__(self, io_profiler=None, root=""):
+        self.username = self._get_principal_name()
```
kuenishi: Why is this moved to before superclass constructor? How is this related to default handlers?

chainerio/_context.py
```diff
@@ -1,116 +1,96 @@
...
-            (handler, self._root, is_URI) = \
-                self.get_handler_by_name(uri_or_handler)
-        assert handler is not None
+            handler, path = self.get_or_create_default_handler(uri_or_handler)
+            handler.root = path
 
-        if self._root:
-            if not handler.isdir(self._root):
-                raise RuntimeError("the URI does not point to a directory")
+            if handler.root:
```
kuenishi: This check must be done before setting `root` to prevent wrong path set to default handlers.

chainerio/filesystems/hdfs.py
```diff
@@ -84,14 +84,27 @@ def _run_klist(use_keytab=False):
...
 
+        self.root = root
         self.nameservice = None
 
+    @property
+    def root(self):
+        return self._root
+
+    # override root
+    @root.setter
```
kuenishi: We don't need this kind of setter as public API.

chainerio/filesystems/hdfs.py
```diff
@@ -84,14 +84,27 @@ def _run_klist(use_keytab=False):
...
         FileSystem.__init__(self, io_profiler, root)
         self.connection = None
         self.type = 'hdfs'
-        self.root = root
-        self.username = self._get_principal_name()
 
+        self.root = root
         self.nameservice = None
 
+    @property
```
kuenishi: As same reason as below on setter, this property method is not needed so far.

chainerio/_context.py
belldandyxtq: resolved in 7c8df6c

belldandyxtq:
Regarding the `.root`, I support the second idea as I did in this PR. There are a few shortages I found with the first approach:
1. `.root` cannot be shared among in different handlers
     e.g. `.root` on `POSIX` usually does not exist on `HDFS`.  Even we cover that in the `set_root`, it would confuse user.
2. `set_root` accepts `IO` handlers, how the `.root` should be is complex when we use one global `root`.
3. How should the `.root` behave when switching handlers and switching back. e.g. Do we restore the old `.root` or set to default.

If you also agree that the second idea is acceptable, I will add the document.

chainerio/filesystems/hdfs.py
belldandyxtq: The `self.username` is used in the setter of `root` at https://github.com/chainer/chainerio/blob/dfeaa0161c4840c68a713fb4b0f928c132e146c3/chainerio/filesystems/hdfs.py#L106, which is overridden at https://github.com/chainer/chainerio/blob/dfeaa0161c4840c68a713fb4b0f928c132e146c3/chainerio/filesystems/hdfs.py#L102 and called in the superclass constructor
https://github.com/chainer/chainerio/blob/dfeaa0161c4840c68a713fb4b0f928c132e146c3/chainerio/io.py#L33

belldandyxtq:
I will rebase the branch after we resolve all the discussion

kuenishi:
Too many conflicts and will seek for other resolution.

PR #103, opened 2019-12-19
Bugfix: test fails when KRB5_KTNAME is not set

belldandyxtq:


kuenishi:
We have to clarify full-spec test environment, like `hdfs` and `klist` is required and several other prerequisites.

PR #105, opened 2020-01-08
check pid in zip

shu65:
This PR solves `ZipContainer` issue in https://github.com/chainer/chainerio/issues/104

### DIFF
--- a/chainerio/containers/zip.py
+++ b/chainerio/containers/zip.py
@@ -1,6 +1,5 @@
 from chainerio.container import Container
 from chainerio.io import open_wrapper
-import copy
 import io
 import logging
 import os
@@ -39,8 +38,6 @@ class ZipContainer(Container):
                 self.zip_file_obj_mode = None
 
         if self.zip_file_obj is None:
-            self.zip_file_obj_pid = os.getpid()
-            self.zip_file_obj_mode = mode
             zip_file = self.base_handler.open(self.base, "rb")
             if isinstance(self.base_handler, ZipContainer) \
                     and sys.version_info < (3, 7, ):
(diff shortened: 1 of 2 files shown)


chainerio/containers/zip.py
```diff
@@ -29,9 +32,15 @@ def _check_zip_file_name(self, base):
         filename, file_extension = os.path.splitext(self.base)
 
     def _open_zip_file(self, mode='r'):
-        mode = mode.replace("b", "")
+        if self.zip_file_obj is not None:
+            if self.zip_file_obj_pid != os.getpid():
+                self.zip_file_obj = None
+                self.zip_file_obj_pid = None
+                self.zip_file_obj_mode = None
```
belldandyxtq: I do not think you need to save the "mode"

tests/container_tests/test_zip_container.py
```diff
@@ -427,3 +433,71 @@ def test_stat(self):
...
+
+            p1 = multiprocessing.Process(target=func)
+            p2 = multiprocessing.Process(target=func)
+            p1.start()
+            p2.start()
+
+            p1.join()
+            p2.join()
+
+            self.assertEqual(p1.exitcode, 0)
```
belldandyxtq: Can you please add a timeout mechanism as the problem here is hanging?

chainerio/containers/zip.py
shu65: `mode` requires to check the mode of previous zip.

chainerio/containers/zip.py
```diff
@@ -30,6 +32,15 @@ def _check_zip_file_name(self, base):
...
+            self._close_zip_file()
+
+        if self.zip_file_obj is not None \
+                and self.zip_file_obj_pid != os.getpid():
+            self.zip_file_obj = None
+            self.zip_file_obj_pid = None
+            self.zip_file_obj_mode = None
 
         if self.zip_file_obj is None:
             zip_file = self.base_handler.open(self.base, "rb")
```
belldandyxtq: Can you please match the mode here as well?

chainerio/containers/zip.py
```diff
@@ -80,12 +94,14 @@ def _wrap_fileobject(self, file_obj: Type['IOBase'],
     def open(self, file_path, mode='r',
              buffering=-1, encoding=None, errors=None,
              newline=None, closefd=True, opener=None):
+        if sys.version_info < (3, 6) and "w" in mode:
+            raise ValueError('Mode w and wb is not supported')
```
belldandyxtq: Can you mention that "w" and "wb" is not supported only in Python < 3.6?

Issue #104, opened 2019-12-25
Avoid accessing implicit shared zip after forking


belldandyxtq:
**Problem statement**
Errors occur when multiple child processes read from an implicitly shared `ZipContainer` that is opened by parents process before forking

**How to reproduce**
```
import chainerio
from multiprocessing import Process

container = chainerio.open_as_container("dummy.zip")
# open a file inside the container to actually open the zip
# the zip will be implicitly shared after forking
with container.open("dummy_data") as f:
    f.read()

def func():
    # accessing the shared container
    with container.open("dummy_data") as f:
        f.read()

p1 = Process(target=func)
p2 = Process(target=func)
p1.start()
p2.start()

p1.join()
p2.join()
```
**Error Message**
```
Process Process-2:
Traceback (most recent call last):
  File "tianqi/.pyenv/versions/3.7.2/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "tianqi/.pyenv/versions/3.7.2/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "test.py", line 10, in func
    with container.open("dummy_data") as f:
  File "tianqi/repository/chainerio/chainerio/io.py", line 20, in wrapper
    errors, newline, closefd, opener)
  File "tianqi/repository/chainerio/chainerio/containers/zip.py", line 88, in open
    nested_file = self.zip_file_obj.open(file_path, "r")
  File "tianqi/.pyenv/versions/3.7.2/lib/python3.7/zipfile.py", line 1488, in open
    raise BadZipFile("Bad magic number for file header")
zipfile.BadZipFile: Bad magic number for file header
Process Process-1:
Traceback (most recent call last):
  File "tianqi/.pyenv/versions/3.7.2/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "tianqi/.pyenv/versions/3.7.2/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "test.py", line 11, in func
    f.read()
  File "tianqi/.pyenv/versions/3.7.2/lib/python3.7/zipfile.py", line 885, in read
    buf += self._read1(self.MAX_N)
  File "tianqi/.pyenv/versions/3.7.2/lib/python3.7/zipfile.py", line 975, in _read1
    data = self._decompressor.decompress(data, n)
[...]

PR #106, opened 2020-01-10
[WIP] Check pid in hdfs

shu65:
This PR solves `HdfsFileSystem` issue in #104


shu65:
The current code stops at these points.
https://github.com/chainer/chainerio/blob/926176af0af635ec3f7e59cb19a25da2616db1d1/chainerio/filesystems/hdfs.py#L131
https://github.com/chainer/chainerio/blob/926176af0af635ec3f7e59cb19a25da2616db1d1/chainerio/filesystems/hdfs.py#L137


We need to consider another way to solve this problem.

PR #110, opened 2020-03-23
WIP: remove krbticket

ksauzz:


belldandyxtq:
Can you also remove the package requirement?
https://github.com/chainer/chainerio/blob/faf6d8bb5ad10526436ede13375646f842e7890c/setup.py#L56

ksauzz:
We decided to leave krbticket as it is for now since krbticket-1.0.4 implemented new mitigation for race condition against credential cache.

PR #112, opened 2020-03-25
Bugfix: list and isdir for non-directory-included zip

belldandyxtq:
This PR fixes a bug which is related to non-directory-included zip, a
kind of zip that contains hierarchical structure but does not contain
the directories themselves. Such kind of zip can be made with "zip -D".

This PR fixes the bug when list or isdir to these "non-exist" directory
in such non-directory-included zip.

We do NOT handle the `exists` in this case as the directories do NOT `exist` **physically**
in the zip, but **logically** `exist`, we want to be honest on what is here and what is not. Also it helps to prevent users from accessing the `stat` of these kind of directory

### DIFF
--- a/chainerio/containers/zip.py
+++ b/chainerio/containers/zip.py
@@ -145,12 +145,16 @@ class ZipContainer(Container):
             given_dir_list = []
 
         if path_or_prefix:
-            if not self.exists(path_or_prefix):
+            if self.exists(path_or_prefix):
+                if not self.isdir(path_or_prefix):
+                    raise NotADirectoryError(
+                        "{} is not a directory".format(path_or_prefix))
+            # check if directories are NOT included in the zip
+            # such kind of zip can be made with "zip -D"
+            elif not any(name.startswith(path_or_prefix + "/")
+                         for name in self.zip_file_obj.namelist()):
                 raise FileNotFoundError(
                     "{} is not found".format(path_or_prefix))
-            elif not self.isdir(path_or_prefix):
-                raise NotADirectoryError(
-                    "{} is not a directory".format(path_or_prefix))
 
         if recursive:
             for name in self.zip_file_obj.namelist():
--- a/chainerio/containers/zip.py
+++ b/chainerio/containers/zip.py
@@ -145,14 +145,13 @@ class ZipContainer(Container):
             given_dir_list = []
 
         if path_or_prefix:
-            if self.exists(path_or_prefix):
-                if not self.isdir(path_or_prefix):
-                    raise NotADirectoryError(
-                        "{} is not a directory".format(path_or_prefix))
-            # check if directories are NOT included in the zip
-            # such kind of zip can be made with "zip -D"
+            if self.exists(path_or_prefix) and not self.isdir(path_or_prefix):
+                raise NotADirectoryError(
+                    "{} is not a directory".format(path_or_prefix))
             elif not any(name.startswith(path_or_prefix + "/")
                          for name in self.zip_file_obj.namelist()):
+                # check if directories are NOT included in the zip
+                # such kind of zip can be made with "zip -D"
                 raise FileNotFoundError(
                     "{} is not found".format(path_or_prefix))
 
--- a/chainerio/containers/zip.py
+++ b/chainerio/containers/zip.py
@@ -200,6 +200,12 @@ class ZipContainer(Container):
                 # is not available before Python 3.6.0
                 return "/" == stat.filename[-1]
         else:
+            file_path = os.path.normpath(file_path)
+            # check if directories are NOT included in the zip
+            if any(name.startswith(file_path + "/")
+                   for name in self.zip_file_obj.namelist()):
+                return True
+
             return False
 
     def mkdir(self, file_path: str, mode=0o777, *args, dir_fd=None):
--- a/tests/container_tests/test_zip_container.py
+++ b/tests/container_tests/test_zip_container.py
@@ -655,12 +655,13 @@ class TestZipHandlerListNoDirectory(unittest.TestCase):
          [NO_DIRECTORY_FILENAME_LIST["testfile2_name"]],
          False],
         # not normalized path
-        ['testdir1//testfile//../',
+        [f'{NO_DIRECTORY_FILENAME_LIST["dir1_name"]}//\
+           {NO_DIRECTORY_FILENAME_LIST["testfile1_name"]}//../',
          [NO_DIRECTORY_FILENAME_LIST["testfile1_name"],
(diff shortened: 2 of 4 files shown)


tests/container_tests/test_zip_container.py
```diff
@@ -559,3 +559,260 @@ def func():
...
+
+        assert stderr == b""
+
+    def tearDown(self):
+        os.chdir(self.pwd)
+        self.tmpdir.cleanup()
+
+    def test_list(self):
+        with self.fs_handler.open_as_container(self.zip_file_name) as handler:
+            cases = [
```
shu65, line 631: could you separate a test function in each case?

In the current version, it cannot be determined in which case failed. Moreover, if one case fails, subsequent cases will not be executed.

tests/container_tests/test_zip_container.py
```diff
@@ -559,3 +559,260 @@ def func():
...
+                zip_generator = handler.list(case['path_or_prefix'],
+                                             recursive=case['recursive'])
+                zip_list = list(zip_generator)
+                print(case["path_or_prefix"], case['expected_list'], zip_list)
+                self.assertEqual(sorted(case['expected_list']),
+                                 sorted(zip_list))
+
+    def test_list_with_errors(self):
+        with self.fs_handler.open_as_container(self.zip_file_name) as handler:
+            cases = [
```
shu65, line 760: Same as above

tests/container_tests/test_zip_container.py
```diff
@@ -559,3 +559,260 @@ def func():
...
+                 "error": FileNotFoundError}]
+            for case in cases:
+                with self.assertRaises(case["error"]):
+                    list(handler.list(case['path_or_prefix']))
+
+                with self.assertRaises(case["error"]):
+                    list(handler.list(case['path_or_prefix'], recursive=True))
+
+    def test_isdir(self):
+        cases = [
```
shu65, line 784: Same as above

chainerio/containers/zip.py
```diff
@@ -145,12 +145,15 @@ def list(self, path_or_prefix: str = "", recursive=False):
             given_dir_list = []
 
         if path_or_prefix:
-            if not self.exists(path_or_prefix):
-                raise FileNotFoundError(
-                    "{} is not found".format(path_or_prefix))
-            elif not self.isdir(path_or_prefix):
+            if self.exists(path_or_prefix) and not self.isdir(path_or_prefix):
```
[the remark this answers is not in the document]
belldandyxtq: The problem here is how we define `exist`. In the common cases, the `dirs` are included in the zip, like this
```
A/
A/file1
A/file2
```

But we can also exclude the `dirs` with `zip -D`. In such cases, the `dir` itself does not `physically exist` in the zip, meaning there is no entries of the `dir`. But the `dir` `exists logically`, in the hierarchy. For example we have a zip which has a `namelist` like
```
A/file1
A/file2
```
The dir `A` itself is not included. Do we return `True` on `exists("A")`? Then how about `stat("A")`?

My option is we be honest on what really `exists` and what does not. That is why I excluded the `exists` in this PR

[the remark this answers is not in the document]
shu65: Thank you for the explanation. 
Since the case  is rare, it seems that `exits()` is OK as this is.

tests/container_tests/test_zip_container.py
```diff
@@ -559,3 +560,258 @@ def func():
...
+         [NO_DIRECTORY_FILENAME_LIST["testfile1_name"],
+          NO_DIRECTORY_FILENAME_LIST["dir2_name"]],
+         False],
+        # problem 2 in issue #66
+        [os.path.join(NO_DIRECTORY_FILENAME_LIST["dir1_name"],
+                      NO_DIRECTORY_FILENAME_LIST["dir2_name"]),
+         [NO_DIRECTORY_FILENAME_LIST["testfile2_name"]],
+         False],
+        # not normalized path
+        ['testdir1//testfile//../',
```
shu65, line 658: Why not use `os.path.join()` and `NO_DIRECTORY_FILENAME_LIST`?

tests/container_tests/test_zip_container.py
```diff
@@ -559,3 +560,258 @@ def func():
...
+          NO_DIRECTORY_FILENAME_LIST["testfile4_name"]],
+         False],
+        # not normalized path beyond root
+        ['//..//',
+         [NO_DIRECTORY_FILENAME_LIST["dir1_name"],
+          NO_DIRECTORY_FILENAME_LIST["dir3_name"],
+          NO_DIRECTORY_FILENAME_LIST["testfile4_name"]],
+         False],
+        # not normalized path beyond root
+        ['testdir1//..//',
```
shu65, line 675: this test case is the same as the case in line 663-667.

tests/container_tests/test_zip_container.py
```diff
@@ -559,3 +560,258 @@ def func():
...
+            self.assertEqual(sorted(expected_list),
+                             sorted(zip_list))
+
+    @parameterized.expand([
+        # non_exist_file
+        ['does_not_exist', FileNotFoundError],
+        # not exist but share the prefix
+        ['t', FileNotFoundError],
+        # broken path
+        ['testdir1//t/', FileNotFoundError],
```
shu65, line 772: How about using `NO_DIRECTORY_FILENAME_LIST["dir1_name"]` insted of `testdir1` ?

tests/container_tests/test_zip_container.py
shu65: If '//' is important, what about this?
`f'{NO_DIRECTORY_FILENAME_LIST["dir1_name"]}//{NO_DIRECTORY_FILENAME_LIST["dir2_name"]}//../'`

belldandyxtq: Sorry, f string is not supported in `python 3.5.2`. I will use the format instead
https://stackoverflow.com/questions/42126794/python-3-returns-invalid-syntax-when-trying-to-perform-string-interpolation

belldandyxtq: https://docs.python.org/3/whatsnew/3.6.html#pep-498-formatted-string-literals

shu65:
Do you fix this test to use 'parameterized' before my review?
https://github.com/chainer/chainerio/blob/dbaf8b04537d29588d53bbdd865720812f4803e2/tests/container_tests/test_zip_container.py#L187

belldandyxtq:
Do you mean "if I want to fix those using `parameterized` in this PR"?

I would like to do it in a separate PR since it is a kind of refactoring and not strongly related to this issue. Is it OK for you?

PR #119, opened 2020-04-23
Rename ChainerIO to PFIO

### LINKED ISSUE #118: Rename to `pfio`
As the `Chainer` has entered the maintenance phrase, we are planing on renaming this library to `pfio`.
But to reduce the porting effort in users code, we can keep the `chainerio` name and let users to import from both name

belldandyxtq:
We rename the ChainerIO to PFIO as `Chainer` has entered the
maintenance phase.

An alias of `chainerio` has created to keep the backward compatibility.
This closes #118


pfio/cache/file_cache.py
```diff
@@ -80,7 +80,7 @@ class FileCache(cache.Cache):
     '''Cache system with local filesystem
 
     Stores cache data in local temporary files, created in
-    ``~/.chainer/chainerio/cache`` by default. Cache data is
+    ``~/.chainer/pfio/cache`` by default. Cache data is
```
belldandyxtq, line 83: Maybe it is better to put under `.pfn/`?

examples/memnn/babi_chainerio.py
```diff
@@ -1,6 +1,6 @@
 import collections
 # CHAINERIO import
```
shu65, line 2: CHAINERIO -> PFIO?

examples/memnn/babi_chainerio.py
```diff
@@ -76,7 +76,7 @@ def read_data(vocab, path):
     data = []
     all_data = []
     # CHAINERIO add
```
shu65, line 78: CHAINERIO -> PFIO?

pfio/cache/file_cache.py
shu65: How about `~/.pfio/cache` ?

Issue #126, opened 2020-06-01
Master fails on style with the latest `flake8`


belldandyxtq:
```
tianqi:pfio$ flake8 pfio
pfio/__init__.py:68:48: F821 undefined name 'IOBase'
pfio/io.py:19:55: F821 undefined name 'IOBase'
pfio/io.py:82:47: F821 undefined name 'IOBase'
pfio/io.py:90:64: F821 undefined name 'IOBase'
pfio/io.py:109:52: F821 undefined name 'IOBase'
pfio/containers/zip.py:119:47: F821 undefined name 'IOBase'
pfio/containers/zip.py:127:64: F821 undefined name 'IOBase'
pfio/filesystems/hdfs.py:180:47: F821 undefined name 'IOBase'
pfio/filesystems/hdfs.py:188:64: F821 undefined name 'IOBase'
tianqi:pfio$ flake8 --version
3.8.2 (mccabe: 0.6.1, pycodestyle: 2.6.0, pyflakes: 2.2.0) CPython 3.7.4 on
Linux
tianqi:pfio$ git branch
  bump_v_1_0_0
  bump_version_0_1_2
* master
  pr_114
  pr_121
  pr_125
  refactor_zip_test
  rename_to_pfio
tianqi:pfio$  
```

PR #128, opened 2020-06-02
MultiprocessFileCache for efficient and easy cache with pytorch dataloader (num_workers>0)

belltailjp:
In case we combine pfio (+HDFS) with PyTorch DataLoader (num_workers>0), due to the following
reasons that are complexly related each other, I noticed that it is hard to make efficient/effective use of pfio FileCache easily.
* With HDFS, setting num_workers>0 is necessary for practical data loading performance
* The current pfio FileCache supports therad-safety but cannot be process-safe
* Currently PyTorch DataLoader is designed to destroy all the worker processes at the end of each epoch

https://github.com/pytorch/pytorch/pull/35795 would solve part of the problems, but being difficult for efficiently setting higher `num_workers` will remain unsolved.

After serious considerations, I figured out that implementing multi-process safe file cache is the simplest way to overcome this issue.
This PR is the introduction of the `MultiprocessFileCache` which imitates the interface and behavior of the existing `FileCache`.

(* modified from the initial implementation)
The new `MultiprocessFileCache` is supposed to be initialized in a single master process and then distributed to its child processes through pickle.
All of these cloned object handle the same cache file(s) created by the master process. The `MultiprocessFileCache` does not keep file descriptors or file objects of the cache files as in the current `FileCache`, but instead it only keeps the filenames. Every time it reads/writes the cache it opens them. This provides process safety.
Cache files remain undeleted as long as the `MultiprocessFileCache` in the *master* process alives, no matter if `MultiprocessFileCache` is destroyed earlier in the *child* process by explicit call of `close`, `exit` nor GC. When the object is destroyed in the master, caches are made sure to be destroyed. This behavior is necessary to prevent the cache is lost in every epoch, when using with PyTorch DataLoader.

A micro-banchmark showed approximately 2x speedup (loading ImageNet zipped val images (including jpeg decode) with `num_workers=16`, cpu=24 setting) in my internal environment.

### DIFF
--- a/pfio/cache/multiprocess_file_cache.py
+++ b/pfio/cache/multiprocess_file_cache.py
@@ -151,6 +151,9 @@ class MultiprocessFileCache(cache.Cache):
         dir (str): The path to the directory to place cache data in
             case home directory is not backed by fast storage device.
 
+        verbose (bool):
+            Print detailed logs of the cache.
+
     '''  # NOQA
 
     def __init__(self, length, do_pickle=False,
@@ -172,6 +175,10 @@ class MultiprocessFileCache(cache.Cache):
         self.data_file = _NoOpenNamedTemporaryFile(self.dir, self._master_pid)
         self.index_file = _NoOpenNamedTemporaryFile(self.dir, self._master_pid)
         index_fd = os.open(self.index_file.name, os.O_RDWR)
+
+        if self.verbose:
+            print('created index file:', self.index_file.name)
+            print('created data file:', self.data_file.name)
 
         try:
             fcntl.flock(index_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
--- a/pfio/cache/multiprocess_file_cache.py
+++ b/pfio/cache/multiprocess_file_cache.py
@@ -13,7 +13,9 @@ import pickle
 class MultiprocessFileCache(cache.Cache):
 
     def __init__(self, length, do_pickle=False,
-                 dir=None, verbose=False):
+                 dir=None, verbose=False,
+                 index_cache_file=None, data_cache_file=None,
+                 cleanup=True):
         self.length = length
         self.do_pickle = do_pickle
         assert self.length > 0
@@ -24,9 +26,19 @@ class MultiprocessFileCache(cache.Cache):
             self.dir = dir
         os.makedirs(self.dir, exist_ok=True)
 
+        self.cleanup = cleanup
         self.closed = False
-        _, self.data_file = tempfile.mkstemp(dir=self.dir)
-        indexfp, self.index_file = tempfile.mkstemp(dir=self.dir)
+        if not data_cache_file:
+            _, self.data_file = tempfile.mkstemp(dir=self.dir)
+        else:
+            self.data_file = data_cache_file
+
+        if not index_cache_file:
+            indexfp, self.index_file = tempfile.mkstemp(dir=self.dir)
+        else:
+            self.index_file = index_cache_file
+            flag = os.O_WRONLY | os.O_TRUNC | os.O_CREAT
+            indexfp = os.open(index_cache_file, flag)
 
         # Initialize index file with index=0, size=-1
         buf = pack('Qq', 0, -1)
@@ -136,9 +148,11 @@ class MultiprocessFileCache(cache.Cache):
     def close(self):
         if not self.closed:
             self.closed = True
-            os.unlink(self.data_file)
-            os.unlink(self.index_file)
+            if self.cleanup:
+                os.unlink(self.data_file)
+                os.unlink(self.index_file)
             self.data_file = None
+            self.index_file = None
 
     @property
     def multiprocess_safe(self) -> bool:
--- a/pfio/cache/multiprocess_file_cache.py
+++ b/pfio/cache/multiprocess_file_cache.py
@@ -156,18 +156,22 @@ class MultiprocessFileCache(cache.Cache):
         self.close()
 
     def __del__(self):
-        # https://github.com/python/cpython/blob/3.8/Lib/tempfile.py#L437
         self.close()
+
+    def _deleter(self, fn, unlink=os.unlink):
+        # https://github.com/python/cpython/blob/3.8/Lib/tempfile.py#L430-L437
+        try:
+            unlink(fn)
+        except FileNotFoundError:
+            # Other worker may have already deleted it
+            pass
 
     def close(self):
         if not self.closed:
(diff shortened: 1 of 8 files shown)


belltailjp:
I have made a few modifications based on an internal discussion
- Use a file for offset and lengths cache (as done in the current FileCache)
- Support what files to use for cache
- Fixed exclusive access control bug
As the result, it allows an intra-node shared cache file (sometimes useful e.g., in imbalanced distributed training setup), and cache file persistency.

In a distributed ImageNet training performance (ResNet50/fp16/8xV100/bs=32 per worker), I confirmed approximately 12% speedup in average, which is very significant.

belldandyxtq:
Is https://github.com/pfnet/pfio/blob/d37b16f55f87fc9217d22e6b0b23bee94b07e8b8/pfio/cache/file_cache.py#L285 helpful in this case? Or you want to also speed up the first epoch data loading

belldandyxtq:
Can you please merge to the latest master? the test fails due to #126 and is fixed by #127

belltailjp:
I think it's indeed very helpful feature.
In the current implementation, the same thing can be achieved by `MultiprocessFileCache(..., index_cache_file='xxxx', data_cache_file='yyy', cleanup=False)`.
`{index|data}_cache_file` arguments are intended to support an intra-node shared cache file, but by combining with `cleanup=False` the caches remain persisted.
If it's better to have the same interface (also better to control timing of persistency process), I think I can implement it, but something equivalent to `{index|data}_cache_file` should still be needed in the initializer for its original purpose, which may be a bit confusing if we have both.
\# Oh, btw it doesn't support what is equivalent to `FileCache.preload`..

belltailjp:
Hi, it's been quite a while, but I update the MultiprocessFileCache implementation. I think it's now ready for review.

* Similar `preserve` and `preload` interface to the existing FileCache.
* Test

pfio/cache/multiprocess_file_cache.py
```diff
@@ -0,0 +1,289 @@
...
+            else:
+                raise ose
+
+    def _put(self, i, data):
+        if self.closed:
+            return
+        assert 0 <= i < self.length
+
+        offset = self.buflen * i
+        index_fd = os.open(self.index_file.name, os.O_RDWR)
```
belldandyxtq, line 188: ditto

pfio/cache/multiprocess_file_cache.py
```diff
@@ -0,0 +1,289 @@
...
+        '''
+
+        index_file = os.path.join(self.dir, '{}.cachei'.format(name))
+        data_file = os.path.join(self.dir, '{}.cached'.format(name))
+
+        if any(os.path.exists(p) for p in (index_file, data_file)):
+            raise ValueError('Specified cache name "{}" already exists in {}'
+                             .format(name, self.dir))
+
+        fd = os.open(self.index_file.name, os.O_RDONLY)
```
belldandyxtq, line 277: ditto

pfio/cache/multiprocess_file_cache.py
```diff
@@ -0,0 +1,289 @@
...
+        else:
+            self.dir = dir
+        os.makedirs(self.dir, exist_ok=True)
+
+        self.closed = False
+        self._frozen = False
+        self._master_pid = os.getpid()
+        self.data_file = _NoOpenNamedTemporaryFile(self.dir, self._master_pid)
+        self.index_file = _NoOpenNamedTemporaryFile(self.dir, self._master_pid)
+        index_fd = os.open(self.index_file.name, os.O_RDWR)
```
belldandyxtq, line 105: Do we close this?

belldandyxtq (changes requested):
Thanks again for your great contribution on this very interesting PR. I have the following questions:
1. Can you please provide some examples on how this cache can be properly used with the `Pytorch DataLoader`? It can be in the docstring.

2. If I understand correctly, the reason why we do not keep the `fd` is to pickle to the subprocesses. But I am concerned that opening/closing the file for every single read usually significantly damage the I/O performance. Is there anyway to reduce the amount of opening/closing? I am not very sure about how exactly the `Pytorch DataLoader` works inside, but is it possible to do things like this:
1. close the fds in master after all the `put`, and set to None
2. fork
3. check the fds in the subprocesses, if it is None, open the fd
In this way we only need one read per subprocesses.

pfio/cache/multiprocess_file_cache.py
```diff
@@ -0,0 +1,289 @@
...
+
+    Arguments:
+        length (int): Length of the cache array.
+
+        do_pickle (bool):
+            Do automatic pickle and unpickle inside the cache.
+
+        dir (str): The path to the directory to place cache data in
+            case home directory is not backed by fast storage device.
+
```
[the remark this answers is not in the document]
belltailjp: It's not actually documented in `FileCache`. Should they be both documented?

belltailjp:
> Can you please provide some examples on how this cache can be properly used with the Pytorch DataLoader? It can be in the docstring.

I added it to the `MultiprocessFileCache` docstring.

> If I understand correctly, the reason why we do not keep the fd is to pickle to the subprocesses. But I am concerned that opening/closing the file for every single read usually significantly damage the I/O performance. Is there anyway to reduce the amount of opening/closing? I am not very sure about how exactly the Pytorch DataLoader works inside, but is it possible to do things like this:

Thanks for the advice! Indeed, it is possible to make sure to open and close the fds only once in each process. I applied this change (though I actually saw no performance improvement in a micro-benchmark).   
Also, dangling fd problem pointed out by your comments are resolved along together.

pfio/cache/multiprocess_file_cache.py
belltailjp: All done in the recent fix :+1:

belldandyxtq (changes requested):

pfio/cache/multiprocess_file_cache.py
```diff
@@ -0,0 +1,364 @@
...
+        '''
+
+        index_file = os.path.join(self.dir, '{}.cachei'.format(name))
+        data_file = os.path.join(self.dir, '{}.cached'.format(name))
+
+        if any(os.path.exists(p) for p in (index_file, data_file)):
+            raise ValueError('Specified cache name "{}" already exists in {}'
+                             .format(name, self.dir))
+
```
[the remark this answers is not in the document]
belltailjp: Right! I forgot it, now I pushed the fix along with a new test case.

belldandyxtq (approved):

PR #132, opened 2020-08-07
Test documentation with python 3.7

belldandyxtq:
numpy >= 1.14 is used when building the documentation. It requires
python >= 3.6, which causes an error in the current CI. This PR fixes
the python version to 3.7 when testing the documentation.
For more details: https://ci.preferred.jp/chainerio/60833/

### DIFF
--- a/.pfnci/test.sh
+++ b/.pfnci/test.sh
@@ -5,6 +5,9 @@ source /root/.bash_docker
 pyenv global 3.5.2 3.6.8 3.7.2
 tox && :
 tox_status=$?
+
+# test doc, needs python >= 3.6
+pyenv global 3.7.2
 pip install .[doc]
 cd docs && make html && :
 sphinx_status=$?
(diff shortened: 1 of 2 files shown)


shu65:
I don't understand the situation.
1. why is numpy >= 1.14 used when building the documentation? 
What version were you using up to now?
2. Why do you need to update the version just by testing the doc? Why doesn't Python for pytest need to update?

belldandyxtq:
Sorry for the lack of explanation.
> why is numpy >= 1.14 used when building the documentation? What version were you using up to now?

The `numpy` is a dependency of `pyarrow`. And the [numpy 1.19](https://numpy.org/devdocs/release/1.19.0-notes.html), which was released on `21 July 2020`, removed the support of `python 3.5`.

>Why do you need to update the version just by testing the doc? Why doesn't Python for pytest need to update?

We are using `tox` to run pytest with python in order of `3.5`, `3.6`, and `3.7`. I am not very sure but I think when we run the test of documentation later, the `numpy` package downloaded from `python 3.7` installed into `python 3.5`. I was trying to find a way to add `--no-cache-dir` to `tox` but it did not work well.

This solution is not the best, but I think it solves the problem some how.

shu65:
Thank you for the explanation.

PR #136, opened 2020-10-08
Cache index error

belltailjp:
Currently, pfio caches check the index of `get` and `put` by the following ways.
- FileCache/MultiprocessFileCache: simple assertion (0 <= i < N)
- NaiveCache: IndexError (i < N; based on the list inside)

In PyTorch, when iterating over a Dataset _without DataLoader_ (which is actually rare case), it doesn't respect the length (`__len__`) of the dataset but instead detects the termination by either IndexError or StopIteration.

```python
>>> import torch
>>>
>>> class Foo1(torch.utils.data.Dataset):
...     def __len__(self):
...         return 3
...     def __getitem__(self, i):
...         print(i)
...         assert 0 <= i < 3
...         return i
... 
>>> class Foo2(torch.utils.data.Dataset):
...     def __len__(self):
...         return 3
...     def __getitem__(self, i):
...         print(i)
...         if 3 <= i:
...             raise IndexError()
...         return i
... 

# Assertion based index check: Since Dataset itself does not respect the __len__, it hits the assertion
>>> foo = Foo1()
>>> for _ in foo:
...     pass
... 
0
1
2
3
  File "<stdin>", line 6, in __getitem__
AssertionError

# IndexError based index check: Dataset catches the IndexError and recognize the termination of the loop.
>>> foo = Foo2()
>>> for _ in foo:
...     pass
... 
0
1
2
3
```

The same thing happens with FileCache and MultiprocessFileCache.
This PR is to make them consistent to this PyTorch behavior, by simply replacing the assertion by IndexError.

Note: Since NaiveCache is based on a list, it can accept the negative index (i<N). In contrast the file based caches requires positive index only (0<= i < N). For now I keep this behavior.

Severity of this PR: Since it is actually rare to iterate over a dataset without DataLoader, so we don't need to consider it as a critical improvement, but sometimes we do so when developing a Dataset class.

### DIFF
--- a/pfio/cache/file_cache.py
+++ b/pfio/cache/file_cache.py
@@ -170,7 +170,10 @@ class FileCache(cache.Cache):
         return data
 
     def _get(self, i):
-        assert i >= 0 and i < self.length
+        if i < 0 or self.length <= i:
+            raise IndexError("index {} out of range ([0, {}])"
+                             .format(i, self.length - 1))
+
         offset = self.buflen * i
         with self.lock.rdlock():
             buf = os.pread(self.indexfp.fileno(), self.buflen, offset)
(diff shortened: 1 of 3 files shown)


kuenishi (approved):

kuenishi:
Now master branch should be clean after #144 , can you try rebasing or merging master, @belltailjp ?

PR #138, opened 2020-11-02
Cache persistency logics to return True/False instead of None/Exception as well as overwrite flag

belltailjp:
The `MultiprocessFileCache` that I introduced in #128 raised inappropriate types of exceptions when preserve/preload are called with an invalid condition, which is inconsistent with `FileCache` behavior.

~* `preserve`: When the destination file already exists, it should raise a `FileExistsError` instead of `ValueError`~
~* `preload`: When the specified cache name does not exist, it should raise a `FileNotFoundError` instead of `ValueError`~

This PR originally provided the fix for the inconsistency. But after a discussion below, now `preserve` and `preload` of both `FileCache` and `MultiprocessFileCache` return a boolean depending on if it succeeded.


pfio/cache/multiprocess_file_cache.py
```diff
@@ -349,8 +350,9 @@ def preserve(self, name):
         data_file = os.path.join(self.dir, '{}.cached'.format(name))
 
         if any(os.path.exists(p) for p in (index_file, data_file)):
-            raise ValueError('Specified cache name "{}" already exists in {}'
-                             .format(name, self.dir))
+            msg = 'Specified cache named "{}" already exists in {}' \
```
belldandyxtq, line 353: I wonder whether we should raise an error in this case. IMO, user might run the same script for multiple times with different parameters, but the cache file name will be the same. If we raise an error here, the job will fail when they do the second run, and user has to remember to clean up the cache each time.

Since it is a read-only cache, do you see any issue if we overwrite it if the same are the same?

belldandyxtq (changes requested):

pfio/cache/multiprocess_file_cache.py
belltailjp: In this PR I intended to have the identical behavior of the original `FileCache` ([os.link called here](https://github.com/pfnet/pfio/blob/1.0.0/pfio/cache/file_cache.py#L303) will raise a `FileExistsError`. Similarly for `preload`, btw).
But I indeed have had the similar concerning to yours. At that moment I considered implementing an 'overwrite=True' option but "always-overwrite" approach looks more handy to me. If you don't mind changing the current behavior I'm happy to update the PR :+1:

belltailjp: This might be related; I have thought of changing the behavior of `preload` when no cache with the specified name exists.
Currently it simply raises `FileNotFoundError` (following the `FileCache` behavior), but similarly to your concerning on repetitive run of a script, it'd be handy if it does nothing when no cache data exists yet.
If a user wants to realize the same thing, currently he/she need to manually check the existence of cache data (which requires several path string manipulations) and call `preload` accordingly.

belltailjp: @belldandyxtq
It's been a while, but I modified the interface of `FileCache` and `MultiprocessFileCache`.
Specifically `preserve` and `preload` return `True` if they succeed and `False` otherwise. Fatal errors such as OSError or invalid use (call preserve/preload in child process) will still raise an Exception, but `FileNotFound` will no longer be thrown.

kuenishi (changes requested):
I think these changes would be nice, but can you describe the behaviour according to your use cases, at the docstring in each case of 

1. `.preserve()` called but the cache file already exists.
2. `.preload()` called but the cache file doesn't exist.

belltailjp:
I've been using this branch in my work for a while since I submitted the PR, then noticed that it can be more convenient by additionally having an `overwrite` option in the `preserve` method, typically when the data format or content to be cached is changed.

I'm thinking of an interface like this:
```python
>>> cache.preserve('cache_name')   # default, overwrite=False (for safety and keep current behavior)
False
>>> cache.preserve('cache_name', overwrite=True)
True
```
It may exceed the original scope of this PR but closely related, so I'd like to hear your opinion if I can make this change in this PR.

kuenishi:
Ah, good idea. The reason it didn't have overwrite option was because there was an assumption that when a cache is preserved, it's complete and there's no need to overwrite a cache file with same name. But cache is cache, and in 1.1.0 cache file may be incomplete and thus overwriting will be often convenient. Go ahead, please.

kuenishi (changes requested):
..and then, this patch conflicts with #145. Please fix the conflicts.

kuenishi (approved):

tests/cache_tests/test_file_cache.py
```diff
@@ -1,83 +1,7 @@
-import os
+from pfio.cache import FileCache
+from pfio.cache import MultiprocessFileCache
```
kuenishi, line 2: This change causes isort fail. Please fix it.
```diff
-from pfio.cache import FileCache
-from pfio.cache import MultiprocessFileCache
 import tempfile
+
+from pfio.cache import FileCache, MultiprocessFileCache
```

pfio/cache/file_cache.py
```diff
@@ -313,19 +332,43 @@ def preserve(self, name):
...
+
         The preserved cache can also be preloaded by
         :class:`~MultiprocessFileCache`.
 
+        Arguments:
+            name (str): Prefix of the preserved file names.
+                ``(name).cachei`` and ``(name).cached`` are created.
+                The files are created in the same directory as the cache
+                (``dir`` option to ``__init__``).
+            overwrite (bool): Overwrite if already exists
```
kuenishi, line 347: Something is wrong around here, which was detected by CI. Please fix it.

```

Warning, treated as error:
/home/kuenishi/src/pfio/pfio/cache/file_cache.py:docstring of pfio.cache.file_cache.FileCache.preserve:21:Definition list ends without a blank line; unexpected unindent.
make: *** [Makefile:19: html] エラー 2
```

kuenishi (changes requested):

kuenishi:
Merging the isort issue is too boring.

PR #141, opened 2020-11-24
Add six to install dependency

kmaehashi:
I failed to install `pfio` on fresh Python environment (`ModuleNotFoundError: No module named 'six'`).
pfio uses `six` but it is not listed in the dependency.


belldandyxtq (approved):
LGTM

PR #143, opened 2021-02-08
Introduce size limit to file caches

belltailjp:
This PR introduces size limitation (in bytes) of cache data file in `FileCache` and `MultiprocessFileCache`.

### Background

Since the file caches are essentially to make a local copy of the entire dataset, they can easily run out of local disk accidentally.
In order to deal with it, the current file caches already handle `errno=ENOSPC` properly, i.e., even when disk full it does not die but just warn without adding the data.

However in some situations this might not be sufficient:
- We basically want to always keep sufficient amount of free space in local (say, GBs) for system stability
- In kubernetes environment, a pod which runs out of local disk will be Evicted (=killed) *before actually getting ENOSPC*

### Functionality

In this PR, the `cache_size_limit` option is added to `FileCache` and `MultiprocessFileCache` respectively.
By default it is set `None`, which means no limitation. So is 0.

In case a valid number is specified, when the total amount of cache data file reaches to the limit, the cache will no longer accept further data addition by changing the cache mode to `frozen`.
The part of cache already built until then can be utilized as usual.

### Known issue

- When multiple `FileCache` or `MultiprocessFileCache` instances exist, there's still a risk for disk exhaustion
    - I think we can make it user's responsibility to set appropriate limitations to each of cache.
- Index file size is not taken into account
    - The file cache has index file which is an LUT for data position as well as data file
    - In this PR, only data file size is considered.
    - Usually size of indices is trivial compared to data (in ImageNet case, index=20MB vs data=34GB)

### DIFF
--- a/pfio/cache/file_cache.py
+++ b/pfio/cache/file_cache.py
@@ -1,3 +1,4 @@
+import numbers
 import os
 from struct import pack, unpack, calcsize
 import threading
@@ -100,16 +101,31 @@ class FileCache(cache.Cache):
         dir (str): The path to the directory to place cache data in
             case home directory is not backed by fast storage device.
 
+        cache_size_limit (None or int): Limitation of the cache size in bytes.
+            If the total amount of cached data reaches the limit,
+            the cache will become frozen and no longer acccept further addition.
+            Data already stored in the cache can be accessed normally.
+            None (default) and 0 is unlimited.
+
         verbose (bool):
             Print detailed logs of the cache.
     '''
 
     def __init__(self, length, multithread_safe=False, do_pickle=False,
-                 dir=None, verbose=False):
+                 dir=None, cache_size_limit=None, verbose=False):
         self._multithread_safe = multithread_safe
         self.length = length
         self.do_pickle = do_pickle
         assert self.length > 0
+
+        if not (cache_size_limit is None or
+                (isinstance(cache_size_limit, numbers.Number) and
+                 0 <= cache_size_limit)):
+            msg = "cache_size_limit has to be either None, zero " \
+                  "(both indicate unlimited) or larger than 0. " \
+                  "{} is specified.".format(cache_size_limit)
+            raise ValueError(msg)
+        self.cache_size_limit = cache_size_limit
 
         if self.multithread_safe:
             self.lock = RWLock()
@@ -186,7 +202,9 @@ class FileCache(cache.Cache):
             return data
 
     def put(self, i, data):
-        assert not self._frozen
+        if self._frozen or self.closed:
+            return False
+
         try:
             if self.do_pickle:
                 data = pickle.dumps(data)
@@ -207,8 +225,12 @@ class FileCache(cache.Cache):
             raise IndexError("index {} out of range ([0, {}])"
                              .format(i, self.length - 1))
 
-        offset = self.buflen * i
+        if self.cache_size_limit:
+            if self.cache_size_limit < (self.pos + len(data)):
+                self._frozen = True
+                return False
 
+        offset = self.buflen * i
         with self.lock.wrlock():
             buf = os.pread(self.indexfp.fileno(), self.buflen, offset)
             (o, l) = unpack('Qq', buf)
--- a/pfio/cache/file_cache.py
+++ b/pfio/cache/file_cache.py
@@ -102,8 +102,8 @@ class FileCache(cache.Cache):
             case home directory is not backed by fast storage device.
 
         cache_size_limit (None or int): Limitation of the cache size in bytes.
-            If the total amount of cached data reaches the limit,
-            the cache will become frozen and no longer acccept further addition.
+            If the total amount of cached data reaches the limit, the cache
+            will become frozen and no longer acccept further addition.
             Data already stored in the cache can be accessed normally.
             None (default) and 0 is unlimited.
 
--- a/pfio/cache/multiprocess_file_cache.py
+++ b/pfio/cache/multiprocess_file_cache.py
@@ -1,5 +1,6 @@
 import errno
 import fcntl
+import numbers
 import os
 import tempfile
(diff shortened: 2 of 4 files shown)


kuenishi:
Thank you for the PR. It's definitely a good idea, but I would like to argue three points.

1. It would be nicer if the behaviour of disk full and cache hits the limit both aligned. I admit that just warning is not enough for today's use cases, which was sufficient in Chainer era, but I want them aligned. I would suggest adding another knob to preserve old behaviour such as `warn_on_full=True`, by default it preserves old behaviour in case of either the cache hits the soft limit or disk full. If it's `False`.
2. I doubt freezing after cache hit limit being really effective. In case of `DistributedDataParallel`, only the partial amount of dataset will be cached, say 1/n (n=number of GPUs) which limits the cache hit rate. Say the task has 8 GPUs, it's cache hit rate is only 12.5% - how much does it improve end-to-end performance?
3. I would suggest implementing this feature in `NaiveCache` too by getting object size with `sys.getsizeof()` . See: https://stackoverflow.com/a/19865746 . It would definitely be useful, in case my question above is addressed, or in case of single process traning.

belltailjp:
For 1 and 3, that makes a lot of sense! I'd like to make changes soon.

For 2, in practice we do not have to worry about it.
In distributed training, we usually split the data into subsets evenly by the number of MPI processes and assign them 1-to-1 using the [`DistributedSampler`](https://pytorch.org/docs/stable/data.html#torch.utils.data.distributed.DistributedSampler). The sampler is created only once at the beginning of training and will be re-used over epochs.
This means that one MPI process and the `FileCache|MultiprocessFileCache` instance inside is made sure to be in charge of only the fixed 1/N subset of the whole dataset during its lifetime.
Therefore, as long as we use the `DistributedSampler` [which is in my understanding common](https://github.com/pytorch/examples/blob/master/imagenet/main.py#L217), the total amount of caches will never change regardless of the number of concurrency. So concurrency won't interfere the hit ratio.

kuenishi:
:+1: on 2. BTW as master CI is somewhat broken, I'll try to fix the broken window and tests will be after that.

kuenishi:
So, I would suggest splitting this idea to two patches. One is to add cache size limit to two classes preserving old behaviour, another is to add option to change the disk/cache-full behaviour to freezing the cache... but there's another point of discussion, is it good to freeze, when the cache is not complete. The meaning of being frozen is the cache is complete and consistent with the original data. Given that our main uses cases are that the dataset is split among distributed processes, it's an irregular case and the application developer doesn't assume it.

A good new behaviour would be choosing among "just warning", "ignore", or "throwing exception (and the program exit)". What do you think?

kuenishi (changes requested):
Just to update PR status. Please re-request review after this gets review-ready again.

belltailjp:
> A good new behaviour would be choosing among "just warning", "ignore", or "throwing exception (and the program exit)". What do you think?

In this PR I implemented the "silent ignore" behavior which I haven't sufficiently clarified why:

The key principle for me is that, when using the cache API in my codes I think it is common to make it as transparent as possible; i.e., the code will work with no problem without cache, and works identically but faster with a cache, like in general memory or OS filesystem caches. Given that, I would expect that my code won't die even if no data can be added anymore **in case** I am aware of it (through explicitly passing the `cache_size_limit` option).
\# Of course, in the case of any serious unexpected problems it should definitely die, though.

For raising an exception, I'd be against it because the original motivation of this PR is to deal with the case where the script is completely killed (by whichever an ENOSPC or eviction in k8s). If the size-limitation API raises an exception, the problem remains just in a similar form.

Warning would be nicer; but I'm afraid of the screen being occupied with the same warning messages; Usually Python suppresses warnings repetition (show just the 1st one), but since this works on process-basis, in our common workloads (MPI and multi-processing) there will still be tons of warning messages filling up.

---

> there's another point of discussion, is it good to freeze, when the cache is not complete.

Right... Honestly speaking, I abused the freeze mechanism just because it is there and handy. I can make some sort of a dedicated flag for the new purpose, though it'll behave identically to the current `._frozen` for now.

The question now is that

> The meaning of being frozen is the cache is complete and consistent with the original data

is this clarified somewhere (or it's fine if you'd define it so now)? If so I won't violate.
Actually `preserve` can be called anytime regardless of cache build progress, which may technically allow existence of an incomplete-but-frozen cache.

belltailjp:
For (3) - implementing `cache_size_limit` in `NaiveCache`, it turns out to be quite difficult...
This is because `sys.getsizeof` does not follow the object reference tree. There is a recipe of calling it recursively,
https://docs.python.org/ja/3/library/sys.html#sys.getsizeof
but it cannot accept arbitrary user-defined types.
There could be some sort of hack (like [this](https://goshippo.com/blog/measure-real-size-any-python-object/) - seems to work quite nicely in pure python world), but the data held outside Python world cannot be detected, which is the case for torch.Tensor.

kuenishi (approved):
I think I understand all of your concerns. I didn't express the definition of being frozen and that was ambiguous in case the cache isn't complete. Let's think that being frozen and complete are both orthogonal state. Maybe it'd be nice if we have a property like `cache.is_complete`, but that should be another discussion. Discussion on behaviour on cache size full can be continued at #138.

So far it's all nice. Thanks!

kuenishi:
@belltailjp Can you fix CI test failure? It's just an annoying flake8 issue.
```

00:06:51.881572 STDOUT 1766] | py39 run-test: commands[1] \| flake8 chainerio pfio |  
-- | -- | --
00:06:52.372922 STDOUT 1766] | pfio/cache/file_cache.py:106:80: E501 line too long (80 > 79 characters) |  
00:06:52.400715 STDOUT 1766] | ERROR: InvocationError for command /repo/.tox/py39/bin/flake8 chainerio pfio (exited with code 1)


```

PR #144, opened 2021-02-08
Fix update python

kuenishi:
Fix the broken window.

- Drop 3.5
- Add 3.8 and 3.9
- Fix pyarrow version to 3.0.0
- Change testing image to GCR-hosted new one based on Ubuntu 20.04


belltailjp (approved):
LGTM! (by local testing with py39)

PR #145, opened 2021-02-10
Single cache file

belltailjp:
Currently file caches (FileCache and MultiprocessFileCache) save the cache information separately into index file and data file.
This PR is to propose to change it to single file.

The new cache file is formatted as:
- index part: (8 bytes sample offset + 8 bytes sample size) * N
- data part: (arbitrary size of binary data) * N

where N is the number of samples to cache.
This is actually completely equivalent to just concatenating the current index and data file.

### More strict cache size limitation (related to #143)
#143 will introduce limitation of cache data, however currently it is limited to data file only.
It is usually OK - as the index files are relatively negligible in most cases (e.g., ImageNet1K case 34GB vs 20MB), but as the number of parallelism grows the index file increases, which becomes relatively non-trivial, but they'll be out of limitation.
By unifying the index and data file into one file, we can constrain the cache size more strictly. 

### Easier handling
In case we want to fully utilize cache `preserve` and `preload` functionality, if the preserved cache is a single file it is easier to handle/manage.

### Performance concern

@kuenishi has pointed out that since in a single-file cache, every time we access to the cache it reads the file twice (to get index and actually data), which may increase chance of conflict in multi-process use.
To be studied.

### DIFF
--- a/pfio/cache/file_cache.py
+++ b/pfio/cache/file_cache.py
@@ -1,3 +1,4 @@
+import errno
 import numbers
 import os
 from struct import pack, unpack, calcsize
@@ -116,7 +117,8 @@ class FileCache(cache.Cache):
         self._multithread_safe = multithread_safe
         self.length = length
         self.do_pickle = do_pickle
-        assert self.length > 0
+        if self.length <= 0 or (2 ** 64) <= self.length:
+            raise ValueError("length has to be between 0 and 2^64")
 
         if not (cache_size_limit is None or
                 (isinstance(cache_size_limit, numbers.Number) and
@@ -132,7 +134,6 @@ class FileCache(cache.Cache):
         else:
             self.lock = DummyLock()
 
-        self.pos = 0
         if dir is None:
             self.dir = _DEFAULT_CACHE_PATH
         else:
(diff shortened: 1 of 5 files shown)


kuenishi:
@belltailjp Tests passed. Is this still WIP?

belltailjp:
I wanted to try some micro-benchmarks to see if there's non-negligible performance regressions due to increased chance of file access concurrency in `MultiprocessFileCache`. :bowing_man:

belltailjp:
https://gist.github.com/belltailjp/eb6a98a799bf1a56bd42132b92db763b#file-result-md
I checked the performance by a simple micro-benchmark.
- Build a cached file with 1MB/sample * 32K samples (32GB total), then measure loading time by 16, 32, 64 and 128 parallel processes.
- Same for 32KB/sample * 1M samples

There are certain amount of fluctuations but I think it's OK to say no significant degradation (nor improvement) in performance.

kuenishi (changes requested):
The code looks good, but please fix the conflict against #143 .

kuenishi (approved):

PR #151, opened 2021-03-03
File access V2 API

kuenishi:
I've drafted a new API to solve existing issues that stem from API design. Partial AWS S3 API support is also added.

(copy from doc)

PFIO v2 API tries to solve the impedance mismatch between different
local filesystem, NFS, and other object storage systems, with a lot
simpler and cleaner code.

It has removed several toplevel functions that seem to be less
important. It turned out that they introduced more complexity than
originally intended, due to the need of the global context. Thus,
functions that depends on the global context such as ``open()``,
``set_root()`` and etc. have been removed in v2 API.

Instead, v2 API provides only two toplevel functions that enable
direct resource access with full URL: ``open_url()`` and
``fs.from_url()``. The former opens a file and returns FileObject. The
latter, creates a ``fs.FS`` object that enable resource access under
the URL. The new class ``fs.FS``, is something close to handler object
in version 1 API. ``fs.FS`` is intended to be as much compatible as
possible, however, it has several differences.

One notable difference is that it has the virtual concept of current
working directory, and thus provides ``subfs()`` method. ``subfs()``
method behaves like ``os.chdir()`` without actually changing current
working directory of the process, but actually returns a *new*
``fs.FS`` object that has different working directory. All resouce
access through the object automatically prepends the working
directory.

V2 API does not provide lazy resouce initialization any more. Instead,
it provides simple wrapper ``recreate_on_fork``, which recreates the
``fs.FS`` object every time the object experiences ``fork(2)``. ``Hdfs``
and ``Zip`` can be wrapped with it, and will be fork-tolerant object.

### DIFF
--- a/pfio/v2/__init__.py
+++ b/pfio/v2/__init__.py
@@ -30,8 +30,7 @@ Or::
 
 from .hdfs import Hdfs, HdfsFileStat  # NOQA
 from .local import Local, LocalFileStat  # NOQA
+from .s3 import S3  # NOQA
 from .zip import Zip, ZipFileStat  # NOQA
-
-# from .s3 import S3
 
 local = Local()
--- a/pfio/v2/__init__.py
+++ b/pfio/v2/__init__.py
@@ -1,0 +1,37 @@
+'''
+fs.FS> interface
+implementations:
+open_fs(URI, container=None|zip) => Local/HDFS/S3/Zip, etc
+- Local
+  - subfs() -> Local
+  - open_zip() -> Zip
+  - open() -> FileObject
+- HDFS
+  - subfs() -> HDFS
+  - open_zip() -> Zip
+  - open() -> FileObject
+- Zip
+  - subfs() -> Zip
+  - open_zip() -> Zip
+  - open() -> FileObject
+- S3 (TBD)
+- GS (TBD)
+
+For example of globally switching backend file systems::
+
+  from pfio.v2 import local as pfio
+
+Or::
+
+  from pfio.v2 import Hdfs
+  pfio = Hdfs()
+
+'''
+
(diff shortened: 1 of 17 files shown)


pfio/v2/__init__.py
```diff
@@ -0,0 +1,36 @@
+'''
+fs.FS> interface
+implementations:
+open_fs(URI, container=None|zip) => Local/HDFS/S3/Zip, etc
+- Local
+  - subfs() -> Local
+  - open_zip() -> Zip
```
belldandyxtq, line 7: By using this name, are we not going to support containers other than zip

pfio/v2/fs.py
```diff
@@ -0,0 +1,381 @@
...
+             closefd: bool = True,
+             opener: Optional[Callable[
+                 [str, int], Any]] = None) -> Type["IOBase"]:
+        raise NotImplementedError()
+
+    def open_zip(self, file_path: str, mode='r') -> Type["Zip"]:  # NOQA
+        from .zip import Zip
+        return Zip(self, file_path, mode)
+
+    def subfs(self, rel_path: str) -> Type["FS"]:
```
belldandyxtq, line 89: Is there any reason we do not use the `chdir` name? Does it support switching from zip to normal directory?

kuenishi: It's rather like chroot than chdir. It doesn't support going up the directory tree or switching the type. See also: https://golang.org/pkg/io/fs/#SubFS

pfio/v2/__init__.py
kuenishi: Do we have any other container format in the scope? I would like being as specific as we need.

PR #159, opened 2021-03-10
Overwrite flush method of ObjectWriter in s3.py

HiroakiMikami:
Some function (e.g., `torch.save`) invoke `flush` method explicitly. But `_ObjectWriter` in s3.py raises `ValueError: ('I/O operation on uninitialized object')` because flush is not supported in S3.

This PR overwrites `flush` method to avoid the above exception.

### DIFF
--- a/pfio/v2/s3.py
+++ b/pfio/v2/s3.py
@@ -62,8 +62,6 @@ class _ObjectWriter(io.BufferedWriter):
 
     def flush(self):
         '''Does nothing
-
-        .. note:: see discussion in ``S3.isdir()``.
         '''
         pass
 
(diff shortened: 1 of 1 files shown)


pfio/v2/s3.py
```diff
@@ -60,6 +60,13 @@ def __init__(self, client, bucket, key, mode, kwargs):
     def write(self, buf):
         self.buf += buf
 
+    def flush(self):
+        '''Does nothing
+
+        .. note:: see discussion in ``S3.isdir()``.
```
kuenishi, line 66: The discussion in `isdir()` doesn't apply here. Please remove this line.

kuenishi (changes requested):
Nit on document

kuenishi (approved):

PR #165, opened 2021-03-11
Fix last_modified attr in S3ObjectStat

HiroakiMikami:
`last_modified` attribute in `S3ObjectStat` is datetime, while Local fs's `last_modified` attribute is floating point value.


kuenishi (changes requested):
`tests/v2_test/test_fs.py::test_smoke()` has a section to test `FileStat` objects. Please move the whole test about being `float` added in this PR  to there.

kuenishi (approved):

PR #175, opened 2021-04-19
Add _checkfork() in S3.open as a foolproof (fix #174)

### LINKED ISSUE #174: Insert _checkfork() into S3 methods to make it foolproof
We often make mistakes; like using `S3` FS objects in PyTorch DataLoader with multi-process mode.

k5342:
This PR adds _checkfork() in S3.open to prevent function call under a multiprocessing situation. Fixes #174.

### DIFF
--- a/tests/v2_tests/test_s3.py
+++ b/tests/v2_tests/test_s3.py
@@ -1,9 +1,10 @@
-import pytest
 import multiprocessing as mp
+
+import pytest
 from moto import mock_s3
 
-from pfio.v2.fs import ForkedError
 from pfio.v2 import S3, from_url
+from pfio.v2.fs import ForkedError
 
 
 @mock_s3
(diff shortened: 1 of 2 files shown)


kuenishi (changes requested):
Please add some test in `tests/v2_tests/test_fs.py` .

kuenishi (approved):

tests/v2_tests/test_s3.py
```diff
@@ -31,3 +34,15 @@ def test_s3():
             assert [] == list(s3.list('base/'))
             assert ['foo.txt'] == list(s3.list('/base'))
             assert ['foo.txt'] == list(s3.list('/base/'))
+
+            def f(s3):
+                try:
+                    s3.open('foo.txt', 'r')
+                except ForkedError:
+                    os._exit(0)
+                os._exit(1)
```
kuenishi, line 43: This code can be written more Pythonic like this:
```python
            def f(s3):
                try:
                    s3.open('foo.txt', 'r')
                except ForkedError:
                    pass
                else:
                    pytest.fail('foo')
```

kuenishi (changes requested):

kuenishi (approved):

PR #176, opened 2021-04-19
Append mock into extras_require for tests

k5342:
The package _mock_ is unlisted in extras_require.


k5342:
_mock_ is included with standard Python library. I will close this PR once and re-submit a new PR with unittest.mock instead.

Issue #214, opened 2021-09-29
Support `O_DIRECT` option in FileCache (and possibly in MultiprocessFileCache)


kuenishi:

kuenishi:
From open(2), 
```
   O_DIRECT
       The O_DIRECT flag may impose alignment restrictions on the length and address of user-space buffers and the file offset of I/Os.  In Linux alignment
       restrictions vary by filesystem and kernel version and might be absent entirely.  However there is currently no filesystem-independent interface for
       an application to discover these restrictions for a given file or filesystem.  Some filesystems provide their own interfaces for doing so, for exam‐
       ple the XFS_IOC_DIOINFO operation in xfsctl(3).
```

kuenishi:
Discussion on OpenZFS is really interesting and worth referring from here. https://github.com/openzfs/zfs/issues/224#issuecomment-124153060

PR #215, opened 2021-09-29
Support O_DIRECT mode for cache files

kuenishi:
Addresses #214 . Fortunately, both multiprocess and non-mp are supported (absolutely only for Linux).


kuenishi:
Note: memory buffer must be aligned with filesystem block boundary. https://stackoverflow.com/a/11809306
Otherwise, it does not work.

PR #219, opened 2021-09-30
Treat prefix ending with "/" as directory

HiroakiMikami:
Currently, S3 does not imitate directory. Thus, we have to branch code based on the filesystem when handling directories with `pfio`.

This PR handles common prefixes that end with "/" as directories in `S3` to increase compatibility between `FileSystem`s.

### DIFF
--- a/pfio/v2/s3.py
+++ b/pfio/v2/s3.py
@@ -30,6 +30,16 @@ class S3ObjectStat(FileStat):
         return False
 
 
+class S3PrefixStat(FileStat):
+    def __init__(self, key):
+        self.filename = key
+        self.last_modified = 0
+        self.size = -1
+
+    def isdir(self):
+        return True
+
+
 class _ObjectReader:
     def __init__(self, client, bucket, key, mode, kwargs):
         self.client = client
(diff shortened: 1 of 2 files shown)


pfio/v2/s3.py
```diff
@@ -398,12 +408,14 @@ def stat(self, path):
             return S3ObjectStat(key, res)
         except ClientError as e:
             if e.response['Error']['Code'] == '404':
+                if self.isdir(path):
+                    return S3PrefixStat(key)
```
kuenishi, line 412: I don't think this is a good design - in cases like kind of a "non-existent" directory (somewhat hard to define), objects in a bucket like
- `foo/bar.txt` and
- `somefile.txt` 

In this case, `s3.isdir('foo')` would be true and also false for `s3.isdir('somefile.txt')`, which is nice, but this code will also return true for `s3.isdir('whatever-nothing')`, which is just nothing.

kuenishi (changes requested):
My suggestion for `S3Stat` is to try `s3.list('somedir')` for imitating 'stat'. It might be heavy, but will behave better.

pfio/v2/s3.py
HiroakiMikami: > but this code will also return true for s3.isdir('whatever-nothing'), which is just nothing.

I think (and I confirmed locally) that `s3.isdir('whatever-nothing')` returns false because this PR handles "common-prefix" as directories. 

on the other hand, I think it's not the best choice in terms of performance. It always calls list-object API when the server returns 404 status code.

kuenishi: Sorry, my bad! I misread the code.

kuenishi (approved):

PR #243, opened 2021-12-22
Add readline method to `_ObjectReader` so S3 file object can be fed to `pickle.load`

belltailjp:
**This PR should be handled after #242**

`pickle.load` requires the received file-object to have `read` *and* `readline` methods.

When opening a .pkl file on S3 using pfio, the type of the returned file-object will be `pfio.v2.s3._ObjectReader` which doesn't implement `readline` method.
As the result, directly passing it to `pickle.load` will result in an error.
```python
>>> with pfio.v2.open_url('s3://<bucket>/testdata.dat', 'rb') as f:
...     pickle.load(f)

TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_13053/1066909613.py in <module>
      1 with pfio.v2.open_url('s3://<bucket>/testdata.dat', 'rb') as f:
----> 2     pickle.load(f)

TypeError: file must have 'read' and 'readline' attributes
```

Seems like `pickle.load` doesn't call the `readline` so I'm not sure why it's required (I cannot even find where in [CPython/Lib/pickle.py](https://github.com/python/cpython/blob/3.9/Lib/pickle.py) raises this TypeError...), but implementing a method without actually doing anything fixes this problem, which is done in this PR.

If it'd be better providing similar behavior to that of normal binary-mode file-like object (returns each line of text but in `bytes`), rather than throwing a `NotImplementedError`, may be I can implement it.
But I think it's not high priority since we don't usually try to get line-by-line strings from a file opened in binary mode.


kuenishi (approved):
/test
LGTM and it's harmless.

PR #245, opened 2021-12-23
Provide 'create' option in each filesystem (v2 API)

belltailjp:
I'd like to have a discussion on adding the `create` option to `from_url`, which I propose a PoC implementation in this PR.

### What is done in this PR
In the current `from_url` (specifically in local and HDFS filesytems), path to a non-existent directory is not accepted.
```python
>>> pfio.v2.from_url("/path/to/non/existent/dir")
ValueError: /path/to/non/existent/directory must be a directory
```

In this PR, I implemented `create` option to  each filesystem:
- `Local` and `HDFS`: When True and if the specified path doesn't exist, it creates the directory
- `S3`: Just ignores
- `Zip`: When True and the specified zip doesn't exist, it raises an error

This allows us to call `from_url` like this:

```python
>>> with pfio.v2.from_url("/path/to/non/existent/dir", create=True) as fs:
... 
>>> os.path.isdir("/path/to/non/existent/dir")
True
```
(Same for HDFS).

### Background

pfio provides so powerful filesystem-agnostic API that we can treat an arbitrary filesytem (`pfio.v2.FS` object) as a portable IO handler to save things, instead of directly using native `open` and `write`.

```python
# A very simple toy example of a native-filesystem dependent data saver.
def save_array(ar, full_path):
    with open(full_path, 'wb') as f:
        np.save(f, ar)

save_array(ar, "/path/to/data.npy")

# A toy filesystem-agnostic saver, which is cool
def save_array(ar, fs, rel_path):
    fs.open(rel_path, "wb") as f:
        np.save(f, ar)

with pfio.v2.from_url("hdfs:///xxx/yyy/dir") as fs:
    save_array(ar, fs, "data.npy")
```

In this way of use, there would be a situation where we want to specify a non-existent directory to `from_url`. In such a case we'd expect `from_url` (or such a saver method) to create the destination if it doesn't exist, but as I mentioned earlier `from_url` doesn't accept it. Specifying an non-existent directory as an output destination (to be created) is a very common idiom that can be seen in various places, so I want to realize the same thing with `from_url`.

\# In that sense, maybe the above toy examples weren't good - in order to save just one file, the proposed `create` option won't sound so much beneficial. When we need to save many files into a directory, the proposed feature becomes more practically useful e.g., training snapshots in deep learning workloads using pytorch-pfn-extras writing abstraction.
https://pytorch-pfn-extras.readthedocs.io/en/latest/reference/generated/pytorch_pfn_extras.training.extensions.snapshot.html

### DIFF
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -4,7 +4,7 @@ import copy
 import os
 import stat
 from abc import abstractmethod
-from io import IOBase, UnsupportedOperation
+from io import IOBase
 from types import TracebackType
 from typing import Any, Callable, Iterator, Optional, Type
 from urllib.parse import urlparse
@@ -335,7 +335,7 @@ def from_url(url: str, **kwargs) -> 'FS':
     def _zip_check_create_not_supported():
         if kwargs.get('create', False):
             msg = '"create" option is not supported for Zip FS.'
-            raise UnsupportedOperation(msg)
+            raise ValueError(msg)
 
     # force_type \ suffix | .zip    | other
     # --------------------+---------+------
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -313,6 +313,8 @@ def from_url(url: str, **kwargs) -> 'FS':
             One of "zip", "hdfs", "s3", or "file", returned
             respectively. Default is ``"file"``.
 
+        create (bool): Create the specified path doesn't exist.
+
     .. note:: Some FS resouces won't be closed when using this
         functionality.
 
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -4,7 +4,7 @@ import copy
 import os
 import stat
 from abc import abstractmethod
-from io import IOBase
+from io import IOBase, UnsupportedOperation
 from types import TracebackType
 from typing import Any, Callable, Iterator, Optional, Type
 from urllib.parse import urlparse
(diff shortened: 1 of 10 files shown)


tests/v2_tests/test_fs.py
```diff
@@ -126,6 +126,48 @@ def test_factory_open():
...
+    with tempfile.TemporaryDirectory() as d:
+        path = os.path.join(d, 'non-existent-directory')
+        with pytest.raises(ValueError):
+            from_url(path)
+
+        from_url(path, create=True)
+        assert os.path.exists(path) and os.path.isdir(path)
+
+    # HDFS
+    hdfs = Hdfs()
```
kuenishi, line 141: Any test that uses `Hdfs()` requires annotation to skip it in CI like in `v2_tests/test_hdfs.py` . I'd recommend moving this section to `test_hdfs.py` to make it clearer.
```python
@unittest.skipIf(shutil.which('hdfs') is None, "HDFS client not installed") 
```

pfio/v2/fs.py
```diff
@@ -330,18 +332,25 @@ def from_url(url: str, **kwargs) -> 'FS':
         if force_type != scheme:
             raise ValueError("URL scheme mismatch with forced type")
 
+    def _zip_check_create_not_supported():
+        if kwargs.get('create', False):
+            msg = '"create" option is not supported for Zip FS.'
+            raise UnsupportedOperation(msg)
```
kuenishi, line 338: I would prefer raising  `ValueError(msg)` indicating that the application using `create` flag with ZIP is wrong.

kuenishi (changes requested):
The overall idea looks great and I like it. But I have a few minor picks.

tests/v2_tests/test_fs.py
belltailjp: Thanks for telling me that :bow::bow:

Right, it should be then moved to `test_hdfs.py` and marked as skip (in CI).
Maybe I can slightly reorganize the tests for `open_url` and `from_url` in `test_fs.py` factories so contributors can write more filesystem-specific precise tests (like in #247) in each `test_xxx.py`.

kuenishi (approved):

Issue #260, opened 2022-01-14
pfio.v2.lazify() may fail in case the PFIO-related context has things that can't pickle


kuenishi:
```python
        pfio.v2.lazify(lambda: pfio.v2.from_url("s3://some-bucket'))
```
Code like this may fail with annoying exception like this:
```
/usr/local/lib/python3.8/multiprocessing/popen_forkserver.py in _launch(self, process_obj)
     45         try:
     46             reduction.dump(prep_data, buf)
---> 47             reduction.dump(process_obj, buf)
     48         finally:
     49             set_spawning_popen(None)

/usr/local/lib/python3.8/multiprocessing/reduction.py in dump(obj, file, protocol)
     58 def dump(obj, file, protocol=None):
     59     '''Replacement for pickle.dump() using ForkingPickler.'''
---> 60     ForkingPickler(file, protocol).dump(obj)
     61 
     62 #

PicklingError: Can't pickle <class 'botocore.client.S3'>: attribute lookup S3 on botocore.client failed
```

Possible fix for this would be use [os.register_at_fork](https://docs.python.org/3/library/os.html#os.register_at_fork). The  function seems to be introduced to solve same problem as we face: https://bugs.python.org/issue6721 .

Caveat: it is since Python 3.7. needs: https://github.com/pfnet/pfio/issues/226

Thanks for original report by @belltailjp

kuenishi:
So the `os.register_at_work` didn't work as expected IMO, and local functions are really hard to pickle and use for tests - I choose another option to resolve the issue in `lazify()` by deprecating it.

PR #266, opened 2022-02-02
Introduce reset_on_fork argument

kuenishi:
To address https://github.com/pfnet/pfio/issues/260 , a new argument `reset_on_fork` is introduced to each FS object constructor. If it is turned on, internal method `_check_fork()` calls a newly introduced abstract method `_reset(self)` . This abstraction enables transparent forking in any situation - be it PyTorch DataLoader or multiprocessing, or "forkserver" enabled.

Classes `S3` and `Hdfs` now do not pickle their internal unpickable, e.g. boto connection or Hdfs connection. If they're forked, they're forced to reset the connection.

It also worked well with PyTorch DataLoader like this:

```python
import time
import os
import multiprocessing

import torch
from pfio.v2 import from_url

class Dataset:
    def __init__(self):
        self.s3 = from_url('s3://bucket/', reset_on_fork=True)

    def __len__(self):
        return 1600

    def __getitem__(self, i):
        with self.s3.open('sl.c', 'rb') as fp:
            return fp.read()

multiprocessing.set_start_method('forkserver', force=True)
#p = multiprocessing.Process()
#p.start()
#p.join()

def main():
    ds = Dataset()
    loader = torch.utils.data.DataLoader(ds, batch_size=8, num_workers=2)
    it = iter(loader)
    data = next(it)
    print(len(data))
    data = next(it)
    print(len(data))
    data = next(it)
    print(type(data))

if __name__ == '__main__':
    main()
```

```sh
$ python dataloader_test.py
8
8
<class 'list'>
```
Note: this change deprecates `pfio.v2.lazify()` .

### DIFF
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -322,8 +322,8 @@ def from_url(url: str, **kwargs) -> 'FS':
     built from it accordingly. The URL path is supposed to
     be a directory for file systems or a path prefix for S3.
 
-    .. warning:: When opening an ``hdfs://...`` URL, be sure
-        about forking context. See: :class:`Hdfs` .
+    .. warning:: When opening an ``hdfs://...`` URL, be sure about
+        forking context. See: :class:`Hdfs` for discussion.
 
     Arguments:
         url (str): A URL string compliant with RFC 1738.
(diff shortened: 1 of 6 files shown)


belltailjp (approved):
[LGTMeow](https://lgtmeow.com)

PR #267, opened 2022-02-09
Fix Cache for numpy

knshnb:
When we use `Cache.get_and_cache()` for `numpy.ndarray`, we get the following error:
```
 ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

This PR fixes the above issue.

### DIFF
--- a/tests/cache_tests/test_cache.py
+++ b/tests/cache_tests/test_cache.py
@@ -83,6 +83,7 @@ def test_cache_numpy(test_class, mt_safe, do_shuffle):
     length = l = 1024
     cache = make_cache(test_class, mt_safe, True, l)
     arr_list = np.random.rand(l, 3, 4)
+
     def getter(i):
         return arr_list[i]
     if do_shuffle:
--- a/tests/cache_tests/test_cache.py
+++ b/tests/cache_tests/test_cache.py
@@ -6,6 +6,7 @@ import tempfile
 
 import numpy as np
 import pytest
+
 from pfio.cache import FileCache, MultiprocessFileCache, NaiveCache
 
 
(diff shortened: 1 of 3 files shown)


kuenishi (approved):

tests/cache_tests/test_cache.py
```diff
@@ -75,6 +75,33 @@ def getter(x):
...
+@pytest.mark.parametrize("test_class", [NaiveCache, FileCache,
+                                        MultiprocessFileCache])
+@pytest.mark.parametrize("mt_safe", [True, False])
+@pytest.mark.parametrize("do_shuffle", [True, False])
+def test_cache_numpy(test_class, mt_safe, do_shuffle):
+    length = l = 1024
+    cache = make_cache(test_class, mt_safe, True, l)
+    arr_list = np.random.rand(l, 3, 4)
+    def getter(i):
+        return arr_list[i]
```
kuenishi, line 87: Please fix the flake8 claim
```
tests/cache_tests/test_cache.py:86:5: E306 expected 1 blank line before a nested definition, found 0
```

kuenishi (changes requested):
Thank you for the patch, I once approved but CI didn't please fix the nit

knshnb:
Thanks! Updated.

tests/cache_tests/test_cache.py
```diff
@@ -4,8 +4,8 @@
 import random
 import tempfile
 
+import numpy as np
 import pytest
-
```
kuenishi, line 8: This line is claimed by isort.

kuenishi (changes requested):
Also, this is the first change that introduces direct dependency on numpy on tests, please add numpy dependency here: https://github.com/pfnet/pfio/blob/2.1.2/setup.py#L55

kuenishi (approved):

PR #270, opened 2022-03-01
Fix `BufferedReader` error caused by opening an empty file in S3 in "rb" mode

### LINKED ISSUE #269: `ValueError: buffer size must be strictly positive` when opening an empty file in S3 with "rb" mode
In case we try to open an empty (zero-bytes) file on S3 in "rb" mode using pfio v2 API with the default buffering option, we get this error.
```python
>>> import pfio; pfio.__version__
'2.2.0'
>>> with pfio.v2.open_url('s3://<bucket>/empty_file.dat', 'rb') as f:
...     pass
... 
  File "/usr/local/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.8/site-packages/pfio/v2/fs.py", line 310, in open_url
    with fs.open(filename, mode) as fp:
  File "/usr/local/lib/python3.8/site-packages/pfio/v2/s3.py", line 407, in open
    obj = io.BufferedReader(obj, buffer_size=bs)
ValueError: buffer size must be strictly positive
```

This is because
https://github.com/pfnet/pfio/blob/2.2.0/pfio/v2/s3.py#L401-L407
```python
    bs = min(obj.content_length, DEFAULT_MAX_BUFFER_SIZE)
```
this will make `bs` to be zero, which is then passed to `BufferedReader` as the buffer size causing the error.

In case the content length of the object to be read is zero, we of course don't need buffering at all, so I think we can simply skip wrapping `obj` by `BufferedReader`.

belltailjp:
Fixes #269

### DIFF
--- a/pfio/v2/s3.py
+++ b/pfio/v2/s3.py
@@ -403,7 +403,7 @@ class S3(FS):
                 bs = min(obj.content_length, DEFAULT_MAX_BUFFER_SIZE)
 
             if 'b' in mode:
-                if self.buffering:
+                if self.buffering and bs != 0:
                     obj = io.BufferedReader(obj, buffer_size=bs)
             else:
                 obj = io.TextIOWrapper(obj)
--- a/tests/v2_tests/test_s3.py
+++ b/tests/v2_tests/test_s3.py
@@ -112,6 +112,19 @@ def test_s3_read(s3_fixture, buffering, reader_type):
             assert not fp.closed
 
 
+def test_empty_file(s3_fixture):
+    with from_url('s3://test-bucket/base',
+                  **s3_fixture.aws_kwargs) as s3:
+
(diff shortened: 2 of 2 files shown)


kuenishi (approved):
Good catch and looks good to me. @belltailjp Do you think you want a 2.2.1 release for this fix?

belltailjp:
To me it's not actually urgent, because in my current workload specifically, empty files being created itself, maybe due to an unknown bug in my code or libraries, is problematic (eventually causes of another error).
So I guess the release timing can be relatively flexible (of course making a release now is also super fine!).

kuenishi:
Thanks for the details. For now, I'll put this for 2.3 schedule, until other issues in high priority come.

PR #282, opened 2022-09-20
HTTPCache feature

y1r:
This PR adds `HTTPCache` feature. `HTTPCache` caches data on HTTP Server. User can use bearer token's path by passing `bearer_token_path` or `PFIO_HTTP_BEARER_TOKEN_PATH` environment variable. The cache will check its bearer token file automatically to support rotation of token.

The HTTP Server should support following operations. This PR also adds sample HTTP server implementation `OnMemoryHTTPServerForTest` in `pfio/testing/__init__.py`.

- GET /url/{integer}
- PUT /url/{integer}


docs/requirements.txt
```diff
@@ -3,3 +3,4 @@ boto3
 deprecation
 sphinx-rtd-theme
 sphinx
+urllib3
```
kuenishi, line 6: I don't think this is necessary for building the document.

pfio/cache/http_cache.py
```diff
@@ -0,0 +1,85 @@
...
+import pickle
+import time
+
+import urllib3
+
+from pfio.cache import Cache
+
+
+class HTTPCache(Cache):
+    def __init__(self, length: int, url: str, bearer_token_path=None, do_pickle=False):
```
kuenishi, line 11: I think this argument spec is misleading to users; We may have to show a clear example of URL such as
```
https://cache.example.com/some/{username}/{dataset-id}/
```
to prevent applications and jobs that use very different dataset or the original data, from interfering each other by overwriting each other's data. In my example, `{username}/{dataset-id}` must be globally identical across the cache system in server side. This is because the access scope depends on the cache systems.

At least the scope or uniqueness requirement of this `url` must be clear in the docstring.

tests/cache_tests/test_cache.py
```diff
@@ -3,105 +3,133 @@
...
-        cache = test_class(length, cache_size_limit=cache_size_limit,
-                           do_pickle=do_pickle,)
-        assert cache.multiprocess_safe
-        assert cache.multithread_safe
-
-    else:
-        assert False
-
-    return cache
+    try:
```
kuenishi, line 22: This try..cache clause is only for HTTP Cache and much wider than the actual scope it needs to be in. To clarify it, please use it inside the `elif test_class == HTTPCache:` clause, e.g.

```python
        elif test_class == HTTPCache:
            assert cache_size_limit is None, \
                "HTTPCache doesn't support cache size limit"

            try:
              httpd, port = make_http_server()
              httpd_thread = Thread(target=httpd.serve_forever)
              httpd_thread.start()
            
              cache = test_class(length, f"http://localhost:{port}/", do_pickle=do_pickle)
              assert cache.multiprocess_safe
              assert cache.multithread_safe
              yield cache
            finally:
              ...
```

Another option to clarify the code is to make the `make_http_server()` as context manager like this:

```python
@contextmanager
def new_make_http_server():
    try:
       httpd, port = make_http_server()
       httpd_thread = Thread(target=httpd.serve_forever)
       httpd_thread.start()
       yield f"http://localhost:{port}/"
    finally:
       ...
```

Then the test would be more simple without large try-cache clause, like
```python
        elif test_class == HTTPCache:
            assert cache_size_limit is None, \
                "HTTPCache doesn't support cache size limit"

            with new_make_cache() as cache_server_url:
              cache = test_class(length, f"http://localhost:{port}/", do_pickle=do_pickle)
              assert cache.multiprocess_safe
              assert cache.multithread_safe
              yield cache
```

pfio/cache/http_cache.py
kuenishi: Also, I would like this class clarified as experimental.

kuenishi (changes requested):
Overall, I like the idea of making the test scope totally under the with statement as well as the design, though I picked a few comments to make the code more clearly readable.

docs/requirements.txt
y1r: Fixed by https://github.com/pfnet/pfio/pull/282/commits/bbe74b6df4d7b484b564dd90c5fb548bbc04d7bf

pfio/cache/http_cache.py
y1r: Thank you for your review. I've added url's uniqueness requirement by https://github.com/pfnet/pfio/pull/282/commits/6e379fda6d2a7e7ab21206a63c36b5e8d4090fb2.

tests/cache_tests/test_cache.py
y1r: I've added `contextmanager` in `make_http_server` to make sure server shutdown will be performed by https://github.com/pfnet/pfio/pull/282/commits/a8e576c484625b8d095004980a0b4a3d564945db. By this, I can avoid using try-catch clause in `make_cache`.

y1r:
@kuenishi Thank you for your review. I've addressed your review comments. Please review this again.

kuenishi (approved):

kuenishi (changes requested):
The CI says the flake8 check is failing: https://ci.preferred.jp/chainerio/112796/#L1203

y1r:
Thanks. I've checked `tox.ini` and fixed for `flake8`, `autopep8`, and `isort`. Could you run CI again ? Also, I've rebased PR.

kuenishi:
/test
I think you can trigger the CI as you're a public member of pfnet, too.

kuenishi (approved):

PR #293, opened 2022-12-19
Refine pfio.cache.HTTPCache

y1r:
This PR refines HTTPCache to support fault-tolerance and multiprocessing with fork-server execution model.

- https://github.com/pfnet/pfio/commit/8ab92dfde2a0bcac909eb298b6bf732df50bc122
  - Reports error (i.e. unexpected HTTP status code or exception) to user instead of stop working
- https://github.com/pfnet/pfio/commit/36b1df82de8a0fe6f4547580c007d0fbeb4ee9b4
  - We can expect that the HTTP backend doesn't return many redirect or errors, so we limit the number of retries to one.
- https://github.com/pfnet/pfio/commit/b138d89c923259dff13837bc572792e991d6f33c
  - We can expect that the HTTP backend processes requests faster I/O, so we shorten the timeout.
- https://github.com/pfnet/pfio/commit/72c9da444218912ccd652881f0cf2067bb382cfe
  - multiprocessing with fork-server execution model requires picklable parameters. The connection pool doesn't support pickle so we re-create the pool if we detected the fork.

### DIFF
--- a/pfio/cache/http_cache.py
+++ b/pfio/cache/http_cache.py
@@ -72,7 +72,7 @@ class HTTPCache(Cache):
         self.do_pickle = do_pickle
 
         # Allow redirect or retry once
-        self.conn = urllib3.poolmanager.PoolManager(retries=1)
+        self.conn = urllib3.poolmanager.PoolManager(retries=1, timeout=3)
 
     def __len__(self):
         return self.length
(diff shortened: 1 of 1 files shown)


kuenishi:
@y1r I've fixed the main CI (just removed flake8), so please rebase and test this pull request again.

Issue #296, opened 2022-12-22
Deprecate `reset_on_fork` flag


kuenishi:
After internal discussion on https://github.com/pfnet/pfio/pull/292, I and @y1r concluded that we don't need that flag any more. If an FS needs reset on fork, just do it and we don't need any flag. Otherwise, just don't reset.

PR #297, opened 2022-12-23
Remove reset_on_fork parameter

y1r:
Fixes https://github.com/pfnet/pfio/issues/296

- Removed `reset_on_fork` parameter from all FS implementations
- ~Removed `ForkedError` because we don't need it now~
- Added `reset_on_fork` parameter check on `from_url` to notify the deprecation

I've found that, in zip + SparseFileCache case, we always use MPCachedWrapper instead of CachedWrapper because we don't have prior information of fork existence by `reset_on_fork` parameter. This may introduce performance decrease by `flock(2)`.

### DIFF
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -3,6 +3,7 @@ import contextlib
 import copy
 import os
 import stat
+import warnings
 from abc import abstractmethod
 from io import IOBase
 from types import TracebackType
@@ -342,7 +343,11 @@ def from_url(url: str, **kwargs) -> 'FS':
 
     '''
     if kwargs.pop('reset_on_fork', None) is not None:
-        print("reset_on_fork is deprecated. PFIO resets on fork by default")
+        warnings.warn(
+            "reset_on_fork is deprecated. PFIO resets on fork by default",
+            category=DeprecationWarning,
+            stacklevel=2
+        )
 
     parsed = urlparse(url)
     if parsed.scheme:
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -67,6 +67,18 @@ class FileStat(abc.ABC):
         return str(self.__str__())
 
 
+class ForkedError(RuntimeError):
+    '''An error class when PFIO found the process forked.
+
+    If an FS object is not "lazy", any object usage detects process
+    fork and raises this ``ForkedError`` as soon as possible at the
+    child process. The parent process may or may not run well,
+    depending on the ``FS`` implementation.
+
+    '''
+    pass
+
+
 class FS(abc.ABC):
     '''FS access abstraction
 
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -67,18 +67,6 @@ class FileStat(abc.ABC):
         return str(self.__str__())
 
 
-class ForkedError(RuntimeError):
-    '''An error class when PFIO found the process forked.
-
-    If an FS object is not "lazy", any object usage detects process
-    fork and raises this ``ForkedError`` as soon as possible at the
-    child process. The parent process may or may not run well,
-    depending on the ``FS`` implementation.
-
-    '''
-    pass
-
-
 class FS(abc.ABC):
     '''FS access abstraction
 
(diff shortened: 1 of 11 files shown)


pfio/v2/fs.py
```diff
@@ -67,28 +67,15 @@ def __repr__(self):
         return str(self.__str__())
 
 
-class ForkedError(RuntimeError):
```
kuenishi, line 70: As stated in the HDFS section, I think this class should be preserved for the real use. Also, it should be preserved for a while in case it's imported in applications' code.

pfio/v2/hdfs.py
```diff
@@ -182,14 +182,10 @@ class Hdfs(FS):
...
     .. warning:: It is strongly discouraged to use :class:`Hdfs` under
-          multiprocessing. Setting ``reset_on_fork=False`` is
-          recommended for HDFS in order not to shoot yourself in the
-          foot, because forking a JVM instance that has a different
-          memory model may cause unexpected behavior. If you do *need*
-          forking, for example, PyTorch DataLoader with multiple
-          workers for performance, it is strongly recommended not to
-          instantiate :class:`Hdfs` before forking. Details are
-          described in PFIO issue #123.  Simple workaround is to set
+          multiprocessing. If you do *need* forking, for example, PyTorch
```
kuenishi, line 185:
```suggestion
          multiprocessing. Once the object detects the process id changed (which means it is forked), the object raises :class:`ForkedError` before doing anything. If you do *need* forking, for example, PyTorch
```

kuenishi (changes requested):
Thank you for the patch. Overall it almost looks quite good, but I want to discuss we'd preserved `ForkedError` from several perspective. The main point is that I want to make it   more fool-proof; raise the error as soon as fork detected, before anything touches JVM which breaks everything and leads to cryptic error messages.

y1r:
@kuenishi Thank you for your review.

https://github.com/pfnet/pfio/pull/297/commits/0ee12550773c2c483ec6a49816db5c03c0f2d4d5 addresses your review comments:
- Added `ForkedError` again
- Checked `multiprocessing.get_start_method()` in the constructor of HDFS to inform users early
- Also, raises `ForkedError` in `_reset()` if start_method is not `forkserver`.

pfio/v2/fs.py
```diff
@@ -338,25 +334,24 @@ def from_url(url: str, **kwargs) -> 'FS':
...
         functionality.
 
     .. note:: Pickling the FS object may or may not work correctly
         depending on the implementation.
 
     '''
-    parsed = urlparse(url)
-    force_type = kwargs.pop('force_type', None)
+    if kwargs.pop('reset_on_fork', None) is not None:
+        print("reset_on_fork is deprecated. PFIO resets on fork by default")
```
kuenishi, line 345: Please use `warnnings.warn(...)` to emphasize it's warning and let the application control it. Maybe [DeprecationWarning](https://docs.python.org/3/library/warnings.html#warnings.warn) suits this case.

kuenishi (changes requested):
Now it looks much better! One nit picked.

pfio/v2/fs.py
y1r: Fixed by https://github.com/pfnet/pfio/pull/297/commits/296804114f1ee9b4dad4210e3b2b5476e64a4a0a

pfio/v2/fs.py
```diff
@@ -338,25 +335,28 @@ def from_url(url: str, **kwargs) -> 'FS':
...
         depending on the implementation.
 
     '''
-    parsed = urlparse(url)
-    force_type = kwargs.pop('force_type', None)
+    if kwargs.pop('reset_on_fork', None) is not None:
+        warnings.warn(
+            "reset_on_fork is deprecated. PFIO resets on fork by default",
+            category=DeprecationWarning,
+            stacklevel=2
```
y1r, line 349: By configuring stacklevel to two, user can see the stacktrace of caller (in user's training script) instead of where warning happen (in PFIO).

kuenishi (approved):

PR #300, opened 2023-01-31
Custom scheme

kuenishi:
When a configuration file `$XDG_CONFIG_HOME/pfio.ini` is placed, it provides an injection point to customize FS resolution from scheme ( `s3` part of `s3://some/path/to/file` ). It lets us configure endpoint, secret, arbitrary configurations that's available at the `kwargs` of each FSes. If you have some s3-compat storage system which is *not* S3, the configuration file would be like:

```ini
[foobar]
scheme = s3
endpoint = https://your-s3-endpoint.example.com
```

And apps will code like this:
```python
  fs = pfio.v2.from_url('foobar://path/to/a')
  with fs.open('filename') as fp:
      fp.read()
```

Possible discussions:
- The configuration object will be static and unchanged during the process. It can be a global configuration variable
- Configuration file name and path (make it configurable with env string?)

### DIFF
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -1,4 +1,5 @@
 import abc
+import configparser
 import contextlib
 import copy
 import os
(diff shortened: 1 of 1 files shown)


kuenishi:
Let us load secrets via environmental variables e.g.

```
[foobar]
scheme = s3
endpoint = https://foobar.example.com/
aws access_key_id = $CUSTOM_ACCESS_KEY
aws_secret_access_key = $CUSTOM_ACCESS_SECRET
```

Other discussions:
- won't do the global configuration variable as like an API
- no need for configurable config path

kuenishi:
Assigned to 2.5.1 instead as it's yet an experimental feature.

PR #314, opened 2023-07-19
Fix Path.glob for compatibility with standard pathlib.Path

niboshi:
Fixed `Path.glob()` for compatibility with standard `pathlib.Path`.

* Made it hierarchy aware. Before `Path.glob("*")` returned all the entries in a recursive manner.
* Fixed so that the base path (on which the glob operation is performed) is included in the returned paths. Before they were relative to the base path.
* Omit trailing `/` for returned directory paths.

Along with that, I did several related changes:

* Refactored internal data of the `Path` class. Specifically, `_root` is merged to `_parts`.
* Added `Path.iterdir()`, `Path.parts` and `Path.root` for compatibility with standard `pathlib.Path`.

### DIFF
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -129,11 +129,13 @@ class FS(abc.ABC):
         elif '..' in rel_path.split(os.path.sep):
             raise RuntimeError("Only subtree is supported")
 
-        sub = copy.copy(self)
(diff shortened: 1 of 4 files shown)


kuenishi (approved):
Thank you so much, @niboshi . I don't even understand which part of my code was wrong, but this patch works perfectly.

PR #315, opened 2023-07-20
Fix pathlib name to follow standard pathlib.Path

belldandyxtq:
This PR matches the `pfio.v2.pathlib.Path.name` with standard pathlib.Path, which only return the file name of the file WITHOUT the directory.

```
import pathlib
import pfio.v2 as pfio

a = pathlib.Path("dir/file.txt")
b = pfio.pathlib.Path("dir/file.txt")

print(f"name from pathlib {a.name}")
print(f"name from pfio {b.name}")
```

output:
```
name from pathlib file.txt
name from pfio dir/file.txt
```

with this PR
```
name from pathlib file.txt
name from pfio file.txt
```

### DIFF
--- a/pfio/v2/pathlib.py
+++ b/pfio/v2/pathlib.py
@@ -37,7 +37,7 @@ class Path:
     @property
     def name(self):
         filename = self._root if not self._parts else self._parts[-1]
-        return os.path.normpath(filename)
+        return os.path.basename(os.path.normpath(filename))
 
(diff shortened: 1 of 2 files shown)


kuenishi:
This bug is also fixed by #314 and this patch conflicts with it. As #314 fixes issues in wider range, I'll take it and closing this patch so far, but thank you.

PR #316, opened 2023-07-24
HTTPCachedFS

y1r:
This PR adds HTTPCachedFS, a wrapper to hook read operations and cache the content with HTTPConnector. HTTPCachedFS wrapper hooks read operations, then try HTTPCache to retrieve its content. The underlying FS will not be called if the data is in the cache. Other operations like write and list will be transferred to the underlying FS.

In order to introduce such a feature, I added `_canonical_name(path)` to FS interface and implementations. `_canonical_name(path)` returns `path` with filesystem information to distinguish Local("abc")._canonical_name("somefile") and S3("abc")._canonical_name("somefile"). S3 and HDFS will encode its endpoint to the path to distinguish multiple endpoints (like AWS's S3 and some S3 compatible storage systems).

Users can enable HTTPCache transparently by following methods:
- Give `http_cache="http://cache-service/"` to `pfio.v2.from_url` method. `from_url` automatically wraps `pfio.v2.FS` to hook all read operations. This will work with local, s3, and zipfs.
  - In the zipfs case, the data in zipfs will be stored in the cache system in a flat manner. It means the url of `somefile` in `hoge.zip` will be `http://{given_url}/hoge.zip/zipfile/somefile`. 
- Or, users can instantiate `HTTPCachedFS` directly

### DIFF
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -354,6 +354,8 @@ def from_url(url: str, **kwargs) -> 'FS':
         create (bool): Create the specified path doesn't exist.
 
         http_cache (str): Prefix url of http cached entries.
+            In the filesystem with http_cache specified, all read access will
+            be hooked and upload its content to the url with the given prefix.
             For details, please refer to :py:class:`pfio.v2.HTTPCachedFS`.
             (experimental feature)
 
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -315,7 +315,7 @@ def open_url(url: str, mode: str = 'r', **kwargs) -> Iterator[IOBase]:
        with open_url("s3://bucket.example.com/path/your-file.txt", 'r') as f:
            f.read()
 
-    .. note:: Some FS resouces won't be closed when using this
+    .. note:: Some FS resources won't be closed when using this
         functionality. See ``from_url`` for keyword arguments.
 
     Returns:
@@ -351,7 +351,11 @@ def from_url(url: str, **kwargs) -> 'FS':
 
         create (bool): Create the specified path doesn't exist.
 
-    .. note:: Some FS resouces won't be closed when using this
+        http_cache (str): Prefix url of http cached entries.
+            For details, please refer to ``HTTPCachedFS``.
+            (experimental feature)
+
+    .. note:: Some FS resources won't be closed when using this
         functionality.
 
     .. note:: Pickling the FS object may or may not work correctly
@@ -376,6 +380,8 @@ def from_url(url: str, **kwargs) -> 'FS':
     if force_type is not None and force_type != "zip":
         if force_type != scheme:
             raise ValueError("URL scheme mismatch with forced type")
+
+    http_cache_prefix = kwargs.pop("http_cache", None)
 
     def _zip_check_create_not_supported():
         if kwargs.get('create', False):
@@ -406,6 +412,10 @@ def from_url(url: str, **kwargs) -> 'FS':
     else:
         dirname = parsed.path
         fs = _from_scheme(scheme, dirname, kwargs, bucket=parsed.netloc)
+
+    if http_cache_prefix is not None:
+        from .http_cache import HTTPCachedFS
+        fs = HTTPCachedFS(http_cache_prefix, fs)
 
     return fs
 
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -298,8 +298,10 @@ class FS(abc.ABC):
         raise NotImplementedError()
 
     @abstractmethod
-    def normpath(self, file_path: str) -> str:
-        """Returns its normpath with protocol and endpoint
+    def _canonical_name(self, file_path: str) -> str:
+        """Returns its canonical name. Canonical name includes its filesystem
+        name, endpoint of filesystem, and file_path to represent its consistent
+        naming. Designed to be used in :py:class:`pfio.v2.HTTPCachedFS`.
         """
         raise NotImplementedError
 
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -352,7 +352,7 @@ def from_url(url: str, **kwargs) -> 'FS':
         create (bool): Create the specified path doesn't exist.
 
         http_cache (str): Prefix url of http cached entries.
-            For details, please refer to ``HTTPCachedFS``.
+            For details, please refer to :py:class:`pfio.v2.HTTPCachedFS`.
             (experimental feature)
 
     .. note:: Some FS resources won't be closed when using this
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -297,6 +297,12 @@ class FS(abc.ABC):
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def normpath(self, file_path: str) -> str:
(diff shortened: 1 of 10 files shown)


pfio/v2/fs.py
```diff
@@ -345,7 +351,11 @@ def from_url(url: str, **kwargs) -> 'FS':
 
         create (bool): Create the specified path doesn't exist.
 
-    .. note:: Some FS resouces won't be closed when using this
+        http_cache (str): Prefix url of http cached entries.
+            For details, please refer to ``HTTPCachedFS``.
```
kuenishi, line 355:
```suggestion
            For details, please refer to :py:class:`pfio.v2.HTTPCachedFS`.
```

pfio/v2/s3.py
```diff
@@ -605,3 +605,13 @@ def remove(self, file_path: str, recursive=False):
         key = _normalize_key(key)
         return self.client.delete_object(Bucket=self.bucket,
                                          Key=key)
+
+    def normpath(self, file_path: str) -> str:
+        path = os.path.join(self.cwd, file_path)
+        norm_path = _normalize_key(path)
+
+        return "s3(endpoint={})/{}/{}".format(
```
kuenishi, line 613: endpoint should also be normalized as well; e.g. trailing slash.

pfio/v2/local.py
```diff
@@ -158,3 +158,6 @@ def glob(self, pattern: str):
         return [
             str(item.relative_to(self.cwd))
             for item in pathlib.Path(self.cwd).glob(pattern)]
+
+    def normpath(self, file_path: str) -> str:
+        return "local" + os.path.normpath(os.path.join(self.cwd, file_path))
```
kuenishi, line 163: `Local` supports relative path. That said, `os.path.join(..)` here can be anything such as `foobar` or `/foo/bar/baz` or `/foo/bar/.././boom/../../../../`. `/foo/bar/../baz` and `/foo/baz` should be treated as same, aren't they?

kuenishi (changes requested):
I imagined a few examples of the so-called `normpath()` output:

```
hdfs(nameservice)/path/to/file
local/path/to/file
s3(endpoint=https://s3compatible.example.com/)/bucket/path/to/file
s3(endpoint=https://s3compatible.example.com/)/bucket/some.zip/zipfile/path/to/file
```

IMO this is not a path defined in POSIX, but is rather close to URL in RFC2718 but it isn't. It could be some canonical identifier - but it's hard to define the spec. I'd want it be not an official method; let's start the name with `_`. If you want to keep the `normpath()` method as official API in part of PFIO, please give it a right name such as `_canonical_name()` or something,  and define the exact input / output spec.

My recommendation is to url-encode the full URL explained in PFIO documentation, while it needs some extension on S3-compatible endpoints: https://github.com/pfnet/pfio/blob/2.6.0/docs/source/design.rst#uri-expression-of-file-paths

On `Zip`, I want to know how to prevent contamination like;

1.  `path/to/file` in `local/foo.zip` named as `local/foo.zip/zipfile/path/to/file`
2. A straight path `local/foo.zip/zipfile/path/to/file`

Anyhow, thank you for the huge contribution.

pfio/v2/local.py
y1r: I think so too, therefore I normalize its path by calling `os.path.normpath()`.

```
Python 3.9.16 (main, Feb 20 2023, 13:34:31) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.path.normpath("/foo/bar/../baz")
'/foo/baz'
>>> os.path.normpath("/foo/baz")
'/foo/baz'
>>> os.path.normpath("/foo/bar/../baz") == os.path.normpath("/foo/baz")
True
```

pfio/v2/http_cache.py
```diff
@@ -0,0 +1,224 @@
...
+    .. note:: This feature is experimental.
+
+    """
+
+    def __init__(self,
+                 url: str,
+                 fs: FS,
+                 bearer_token_path: Optional[str] = None):
+        super().__init__()
+        self.fs = fs
```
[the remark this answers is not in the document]
y1r: Fixed by https://github.com/pfnet/pfio/pull/316/commits/f6f3b52ef82bc95a67259025021ed75a0c421a92

pfio/v2/http_cache.py
```diff
@@ -0,0 +1,224 @@
+import io
+from types import TracebackType
+from typing import Any, Iterator, Optional, Type, Union
+
+from pfio.cache import HTTPConnector
+
+from .fs import FS, FileStat
+
+
+class HTTPCachedFS(FS):
```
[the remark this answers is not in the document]
y1r: Fixed by https://github.com/pfnet/pfio/pull/316/commits/f6f3b52ef82bc95a67259025021ed75a0c421a92

pfio/v2/http_cache.py
```diff
@@ -0,0 +1,224 @@
...
+        data = self.conn.get(self.cache_path)
+        if data is not None:
+            self.whole_file = data
+            return
+
+        with self.fs.open(self.file_path,
+                          *self.open_args, **self.open_kwargs) as f:
+            self.whole_file = f.read(-1)
+
+        if 1024 * 1024 * 1024 <= len(self.whole_file):
```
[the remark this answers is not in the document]
y1r: Fixed by https://github.com/pfnet/pfio/pull/316/commits/ab79d2fe45eb91247049b50839a9eecd7db37afd

pfio/v2/s3.py
y1r: Fixed by https://github.com/pfnet/pfio/pull/316/commits/0c5974c5d3d264cf45e125b0de2177fcdfbba7ea. Parsed endpoint and use its hostname only

y1r:
Thank you for your detailed review! I renamed `normpath` to `_canonical_name`, and it now returns url-like string such as `scheme://endpoint/bucket/key_name`. Also, `zipfile` is too broad name as reserved name, thus I renamed it to `pfio-zipfs` to avoid namespace pollution.

pfio/v2/local.py
```diff
@@ -158,3 +158,6 @@ def glob(self, pattern: str):
         return [
             str(item.relative_to(self.cwd))
             for item in pathlib.Path(self.cwd).glob(pattern)]
+
+    def _canonical_name(self, file_path: str) -> str:
+        return "local:/" + os.path.normpath(os.path.join(self.cwd, file_path))
```
kuenishi, line 163: Default scheme of local filesystem is `file:/`.

```suggestion
        return "file:/" + os.path.normpath(os.path.join(self.cwd, file_path))
```

See also:
- https://pfio.readthedocs.io/en/latest/design.html#uri-expression-of-file-paths
- `fs._from_scheme()`
- https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

kuenishi (changes requested):

pfio/v2/local.py
y1r: Thank you for your review. Fixed by https://github.com/pfnet/pfio/pull/316/commits/4ad0497bd4abbbd60ca308fa99bfec3bb5d4b1e4

kuenishi (approved):

PR #317, opened 2023-07-25
Faster HTTPConnector

y1r:
I profiled HTTPConnector's performance by using [cProfile](https://docs.python.org/3/library/profile.html) and [snakeviz](https://jiffyclub.github.io/snakeviz/). I found that urllib3's HTTP request is not optimized for sending requests to the same host.

Previous: 
image
The given url is parsed many times to get appropriate connection from the connection pool. The url parse is actually just execution of regexp, but from the perspective of small file use case (latency-neck), it should not be happened many times.

This PR:
image
I re-implemented connection pool for sending requests to the same host. By this, we can see the performance improvement (we're using more time in `recvinto` relatively, so it means the latency is reduced). Also, we don't have to use urllib3 anymore.

Note: I don't implement clean up of connection pool, but if user don't use many hosts to request, it will not be a problem.

### DIFF
--- a/pfio/cache/http_cache.py
+++ b/pfio/cache/http_cache.py
@@ -1,11 +1,12 @@
+import collections
+import contextlib
+import http.client
 import logging
 import os
 import pickle
 import time
-from typing import Any, Optional
-
-import urllib3
-import urllib3.exceptions
+import urllib.parse
+from typing import Any, Dict, Generator, List, Optional
 
 from pfio.cache import Cache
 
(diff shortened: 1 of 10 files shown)


pfio/cache/http_cache.py
```diff
@@ -33,36 +34,66 @@ def __setstate__(self, state):
...
 
 class HTTPConnector(object):
     def __init__(self,
                  url: str,
                  bearer_token_path: Optional[str] = None,
-                 retries: int = 1,
                  timeout: int = 3):
-        if url.endswith("/"):
-            self.url = url
+        parsed = urllib.parse.urlparse(url)
```
y1r, line 76: url parsing only happened in the constructor. previously it happened in get/put.

PR #318, opened 2023-07-27
Rewrite HTTPConnector by http.client

y1r:
Based on https://github.com/pfnet/pfio/pull/317 work, but cherry-picked

I profiled HTTPConnector's performance by using [cProfile](https://docs.python.org/3/library/profile.html) and [snakeviz](https://jiffyclub.github.io/snakeviz/). I found that urllib3's HTTP request is not optimized for sending requests to the same host.

Previous: 
image
The given url is parsed many times to get appropriate connection from the connection pool. The url parse is actually just execution of regexp, but from the perspective of small file use case (latency-neck), it should not be happened many times.

This PR:
image
I re-implemented connection pool for sending requests to the same host. By this, we can see the performance improvement (we're using more time in `recvinto` relatively, so it means the latency is reduced). Also, we don't have to use urllib3 anymore.

Note: I don't implement clean up of connection pool, but if user don't use many hosts to request, it will not be a problem.

### DIFF
--- a/pfio/cache/http_cache.py
+++ b/pfio/cache/http_cache.py
@@ -39,6 +39,7 @@ class _ConnectionPool(object):
             Generator[http.client.HTTPConnection, None, None]:
         if self.is_forked:
             self.conn = collections.defaultdict(list)
+            self.pid = os.getpid()
 
         conn: http.client.HTTPConnection
         try:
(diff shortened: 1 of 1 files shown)


y1r:
CI problem is related to https://github.com/pfnet/pfio/pull/320

y1r:
Rebased on https://github.com/pfnet/pfio/pull/320

kuenishi (approved):

PR #321, opened 2023-07-31
Add function interface to add custom scheme

niboshi:
pfio has the ability to configure additional custom schemes via an ini file (https://pfio.readthedocs.io/en/latest/design.html#custom-scheme). However, there are cases where we prefer not to rely on the environment and instead configure custom schemes explicitly.

`pfio.v2.config.add_custom_scheme()` in this PR would allow programs to configure custom schemes prior to using other pfio features without relying on an environment file.

### DIFF
--- a/pfio/v2/config.py
+++ b/pfio/v2/config.py
@@ -1,0 +1,46 @@
+import configparser
+import os
+from typing import Dict, Optional
+
+
+def _default_config_file():
+    path = os.getenv('PFIO_CONFIG_PATH')
+    if path:
+        return path
+
+    basedir = os.getenv('XDG_CONFIG_HOME')
+    if not basedir:
+        basedir = os.path.join(os.getenv('HOME'), ".config")
+
+    return os.path.join(basedir, "pfio.ini")
+
+
+def _reload_config():
+    global _config
+    config = configparser.ConfigParser()
+    configfile = _default_config_file()
+    config.read(configfile)
+    _config = config
+
+
+def add_custom_scheme(name: str, scheme: str, data: Optional[Dict[str, str]] = None):
+    if data is None:
+        data = {}
+    else:
+        data = data.copy()
+
+    data["scheme"] = scheme
+    _config[name] = data
+
+
+def get_custom_scheme(name: str) -> Optional[Dict[str, str]]:
+    try:
+        return dict(_config[name])
+    except KeyError:
+        return None
+
+
+_config = None
+
+
+_reload_config()
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -1,5 +1,4 @@
 import abc
-import configparser
 import contextlib
 import copy
 import os
@@ -13,6 +12,7 @@ from urllib.parse import urlparse
 
 from deprecation import deprecated
 
+from pfio.v2 import config
 from pfio.version import __version__  # NOQA
 
 
@@ -404,44 +404,13 @@ def from_url(url: str, **kwargs) -> 'FS':
     return fs
 
 
-def _default_config_file():
-    path = os.getenv('PFIO_CONFIG_PATH')
-    if path:
-        return path
-
-    basedir = os.getenv('XDG_CONFIG_HOME')
-    if not basedir:
-        basedir = os.path.join(os.getenv('HOME'), ".config")
-
-    return os.path.join(basedir, "pfio.ini")
-
-
-class _CustomScheme:
-    conf = None
-
-    @staticmethod
-    def config(scheme):
-        if _CustomScheme.conf is None:
-            _CustomScheme.load_config()
-
-        if scheme in _CustomScheme.conf:
-            return dict(_CustomScheme.conf[scheme])
-
-    @staticmethod
-    def load_config():
-        config = configparser.ConfigParser()
-        configfile = _default_config_file()
-        config.read(configfile)
-        _CustomScheme.conf = config
-
-
 def _from_scheme(scheme, dirname, kwargs, bucket=None):
     known_scheme = ['file', 'hdfs', 's3']
 
     # Custom scheme; using configparser for older Python. Will
     # update to toml in Python 3.11 once 3.10 is in the end.
     if scheme not in known_scheme:
-        config_dict = _CustomScheme.config(scheme)
+        config_dict = config.get_custom_scheme(scheme)
         if config_dict is not None:
             scheme = config_dict.pop('scheme')  # Get the real scheme
             # Custom scheme expected here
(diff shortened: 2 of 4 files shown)


niboshi:
Is this PR responsible for these errors? 

>tests/v2_tests/test_fs.py:87:16: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
>tests/v2_tests/test_s3_zip.py:29:48: E231 missing whitespace after ','
>tests/v2_tests/test_s3_zip.py:160:48: E231 missing whitespace after ','

pfio/v2/config.py
```diff
@@ -0,0 +1,50 @@
...
+    global _config
+    config = configparser.ConfigParser()
+    configfile = _default_config_file()
+    config.read(configfile)
+    _config = config
+
+
+def add_custom_scheme(
+    name: str,
+    scheme: str,
```
kuenishi, line 28: I don't want to have `scheme` independent input but just have built in a flat dictionary. Thus this should not be just for scheme, but for general configurations and shall be named as `add_config(..)` or `add_config_section(..)` and such. The config file is not just for custom scheme, but for more general configuration (which is not yet to be).

pfio/v2/config.py
```diff
@@ -0,0 +1,50 @@
...
+    data: Optional[Dict[str, str]] = None,
+) -> None:
```
kuenishi, line 30: As custom scheme is still experimental, this function should be mentioned explicitly experimental and documented somehow in docs as well.

pfio/v2/config.py
```diff
@@ -0,0 +1,50 @@
...
+    if data is None:
+        data = {}
+    else:
+        data = data.copy()
+
+    data["scheme"] = scheme
+    _config[name] = data
+
+
+def get_custom_scheme(name: str) -> Optional[Dict[str, str]]:
```
kuenishi, line 40: Ditto

pfio/v2/config.py
```diff
@@ -0,0 +1,50 @@
...
+    try:
+        return dict(_config[name])
```
kuenishi, line 42:
```suggestion
        return dict(_config).get(name)
```
I'd prefer flat style to try-except handling.

pfio/v2/config.py
```diff
@@ -0,0 +1,50 @@
...
+    except KeyError:
+        return None
+
+
+_config = None
+
+
+_reload_config()
```
kuenishi, line 50: I don't want to change the timing of loading configuration file. Currently it was read every time opening any `FS`, but with this line it'll be changed just once on global startup. How about keeping dynamically added dict separately, and merge it with configs from the file every time?

pfio/v2/fs.py
```diff
@@ -462,7 +431,7 @@ def _from_scheme(scheme, dirname, kwargs, bucket=None):
         from .s3 import S3
         fs = S3(bucket=bucket, prefix=dirname, **kwargs)
     else:
-        raise RuntimeError("bug: scheme '{}' is not supported".format(scheme))
+        raise RuntimeError("scheme '{}' is not supported".format(scheme))
```
kuenishi, line 434: Maybe reaching this line has not been a bug, but then "not supported" is too strong regarding the spec schemes can be dynamically added. Maybe change it as "not defined"?

kuenishi (changes requested):
Thank you for the contribution. The CI failure was fixed by https://github.com/pfnet/pfio/pull/320 so please rebase against the current master to fix the failure.

Add a few comments important as this patch adds new API to the library so I'd like to be a bit cautious.

pfio/v2/config.py
niboshi: I thought this config mechanism is just for custom schemes, because a config key is a scheme name itself.
My assumption is that more fundamental refactoring of low-level config mechanism would be needed to support general config entries. So I think it's better to provide a higher-level `add_custom_scheme` function here.
(I feel no problem about merging `scheme` and `data` arguments though)

How do you think?
That being said, I'm fine with either ways.

pfio/v2/config.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def get_custom_scheme(name: str) -> Optional[Dict[str, str]]:
+    try:
+        return dict(_config[name])
+    except KeyError:
+        return None
+
+
+_config = None
```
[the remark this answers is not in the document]
niboshi: `_CustomScheme` was never instantiated and used just as a namespace.
In this PR, a separate `config` module is created, which works as an alternative to it, so I thought `_CustomScheme` is no longer needed.

pfio/v2/config.py
niboshi: Fixed in b34fad7cc246354da22827d8870a918f5d9d061d. (you can't just use `.get` because `dict()` is applied after that)

pfio/v2/fs.py
niboshi: Fixed in 13319b2638a4a5124e9cdab5001324cb2a90134f.

pfio/v2/config.py
kuenishi: Yes, I should have expressed my intention more clearly. I understand your reasoning and it's quite correct. Making the section names as custom scheme names was too narrow idea. Maybe we can agree later on how much fundamental refactoring needed for general config entries. Once we got it, this `add_custom_scheme()` can be changed to a shallow wrapper of it.

So my conclusion here would be that, my request is just to add docs here. No need to change function names.

kuenishi: No, you don't have to do that, but +1 to check the `scheme`.

niboshi: Thank you. Added docstrings: 61dd14b94eeeaab45ee8a2beee8d2e57f6852a9a

pfio/v2/config.py
niboshi: Added docstrings: 61dd14b94eeeaab45ee8a2beee8d2e57f6852a9a

pfio/v2/config.py
niboshi: Added docstrings: 61dd14b94eeeaab45ee8a2beee8d2e57f6852a9a

pfio/v2/config.py
niboshi: Fixed for lazy loading: a016f14ff6f3f4cce13e7a6c1605718fbd6a16e3

But I just realized I didn't read your comment carefully.
Could you elaborate a bit more?
(Do you mean `get_custom_scheme()` would look up from both `_configs` and the separate dict? I don't get why the separate dict would be needed.)

niboshi:
Thank you.
I have a question on the comment about loading: https://github.com/pfnet/pfio/pull/321#discussion_r1311020044
Please check it :bow:

kuenishi (approved):

pfio/v2/config.py
kuenishi: I thought that both the configuration file and the global dict `data` could be both updated anytime during the process lifetime. But maybe that'd be too much and ineffective.

PR #323, opened 2023-08-24
Add __repr__ to FS classes

niboshi:
It would be convenient if some information was included in `repr`.

```
<pfio.v2.local.Local object at 0x7f6b09c7d120>
```
↓
```
pfio.v2.local.Local(cwd='/')
```

### DIFF
--- a/pfio/v2/_utils.py
+++ b/pfio/v2/_utils.py
@@ -1,0 +1,6 @@
+from typing import Any, Dict, Type
+
+
+def format_repr(cls: Type, data: Dict[str, Any]) -> str:
+    data_str = ", ".join(f"{name}={value!r}" for name, value in data.items())
+    return f"{cls.__module__}.{cls.__name__}({data_str})"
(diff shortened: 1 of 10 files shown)


pfio/v2/_utils.py
```diff
@@ -0,0 +1,6 @@
+from typing import Any, Dict, Type
+
+
+def format_repr(cls: Type, data: Dict[str, Any]) -> str:
```
kuenishi, line 4: I want this function included in the tail of `fs.py` to obey the no-util principle.

kuenishi (changes requested):
Thank you, that's wonderful suggestion and this is what I really wanted. I want to merge it so hard but also I want to keep the code under the principle of "no util" to prevent it being source of mess in the near future. c.f. https://mykeels.medium.com/utils-are-a-blackhole-c8320968325b
https://dev.to/dvddpl/utils-files-are-not-so-useful-and-helper-classes-are-not-so-helpful-1kfn

pfio/v2/_utils.py
niboshi: Thank you for the commend. Fixed: 314682840a26a88c69c1f8f208473367772c6320

kuenishi (approved):
:+1:

PR #328, opened 2023-10-05
S3 support for large files

alexisVallet:
Current code on `master` fails to handle large files on AWS S3 in 2 ways:
1. When performing multi-part uploading, the md5 given to `upload_part` is hexadecimal, but base64 is expected.
2. When calling `S3.rename` with a source object larger than 5GB, it fails due to limitations of `copy_object`.

While issue 1. is a straightforward fix. Solving issue 2. requires using multi-part copying, which (in my understanding) makes `rename` no longer atomic for large files. Considering `rename` is often used for its atomicity, I'm not sure whether this solution is best, and would be happy to get feedback.

### DIFF
--- a/pfio/v2/s3.py
+++ b/pfio/v2/s3.py
@@ -595,18 +595,21 @@ class S3(FS):
 
         '''
         self._checkfork()
-        source = {
-            'Bucket': self.bucket,
-            'Key': _normalize_key(os.path.join(self.cwd, src)),
-        }
+        src_key = _normalize_key(os.path.join(self.cwd, src))
         dst = os.path.join(self.cwd, dst)
         dst = _normalize_key(dst)
-        res = self.client.copy_object(Bucket=self.bucket,
-                                      CopySource=source,
-                                      Key=dst)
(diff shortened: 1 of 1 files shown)


pfio/v2/s3.py
```diff
@@ -193,11 +194,10 @@ def _flush(self):
 
         data = self.buf.getvalue()
         if 'b' in self.mode:
-            md5 = hashlib.md5(data).hexdigest()
+            md5 = base64.b64encode(hashlib.md5(data).digest()).decode()
         else:
-            md5 = hashlib.md5(data.encode()).hexdigest()
+            md5 = base64.b64encode(hashlib.md5(data.encode()).digest()).decode()
```
kuenishi, line 199: This line yields CI error: `pfio/v2/s3.py:199:80: E501 line too long (80 > 79 characters)` by flake8.
Please add `# NOQA` or newline.

alexisVallet: Thanks! I solved the linting issues and now the CI is passing.

kuenishi (approved):
Great work! Thank you so much.

PR #331, opened 2024-01-12
Path: more compatibility with pathlib.Path

yhrkw:
This PR improves compatibility between `pfio.v2.pathlib.Path` and `pathlib.Path`.

The key differences between this PR and current impl./`pathlib.Path` as follows.

- many methods raise `NotImplementedError` because they require unsupported features of `FS`
- provides `scheme` property for `as_uri` (will move to `FS`)
- `as_uri` returns `file`, `s3`, `hdfs` or custom scheme URI dependent on `fs` and `scheme`
- `stat` returns `pfio.v2.FileStat` instead of `os.stat_result`

### DIFF
--- a/pfio/v2/pathlib.py
+++ b/pfio/v2/pathlib.py
@@ -1,271 +1,756 @@
-import fnmatch
-import os
-from pathlib import PurePath
+import functools
+import uuid
+from fnmatch import fnmatch, fnmatchcase
+from io import IOBase
+from os import PathLike
+from pathlib import PurePosixPath
+from posixpath import join as joinpath
+from posixpath import normpath
+from sys import version_info as python_version_info
+from typing import (Callable, Iterator, List, Optional, Sequence, Set, Tuple,
+                    Type, TypeVar, Union)
+from urllib.parse import ParseResult, urlunparse
 
-from .fs import FS
+from .fs import FS, FileStat
+from .hdfs import Hdfs
 from .local import Local
+from .s3 import S3
+from .zip import Zip
+
+SelfPurePathType = TypeVar("SelfPurePathType", bound="PurePath")
+SelfPathType = TypeVar("SelfPathType", bound="Path")
 
 
-class Path:
-    sep = '/'
+@functools.lru_cache(maxsize=16)
+def _has_directory_feature(fs: FS) -> bool:
+    if isinstance(fs, S3):
+        # NOTE: Apache Ozone supports supports directory or folder.
+        # refers to `ozone.om.enable.filesystem.paths` property.
+        dpath = f"__pfio_check_{str(uuid.uuid4())[:8]}"
+        upath = f"{dpath}/__tmp"
 
-    def __init__(self, *args, fs=None, root=None):
-        '''
+        with fs.open(upath, mode="wb") as f:
+            f.write(b"")
+        fs.remove(upath)
 
-        Empty argument indicates current directory in fs, fs.cwd
-
-        '''
-        self._fs = fs if fs else Local()
-        assert isinstance(self._fs, FS)
-
-        sep = self.sep
-        if root:
-            if not root.startswith(self.sep):
-                raise ValueError("root must start with a separator.")
-            if root == self.sep:
-                args = [self.sep] + args
-            else:
-                args = [self.sep, root[len(self.sep):]] + args
-
-        # Construct parts
-        parts = []
-        for argv in args:
-            p = argv
-            if p.startswith(sep):
-                parts = [sep]
-                p = p[len(sep):]
-            if len(p) > 0:
-                parts += [q for q in p.split(sep) if q not in ("", ".")]
-
-        # Check
-        assert "" not in parts
-        if parts and parts[0] == self.sep:
-            assert all(self.sep not in p for p in parts[1:])
-        else:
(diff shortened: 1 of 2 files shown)


pfio/v2/pathlib.py
```diff
@@ -1,271 +1,756 @@
...
+    # ---------------------------------------
+    # pathlib.Path compatible classmethods
+    # ---------------------------------------
 
-    def read_text(self):
-        with self.open() as fp:
-            return fp.read()
+    @classmethod
+    def cwd(cls: Type[SelfPathType]) -> SelfPathType:
+        raise _not_supported()
```
kuenishi, line 373: Some FSes support .cwd including `pfio.v2.Local`. I want underlying FSes handle this method call (if it doesn't implement cwd, then let it throw unsupported exception. Oh, it has been never implemented. Nevermind..

pfio/v2/pathlib.py
```diff
@@ -1,271 +1,756 @@
...
 
-    def with_suffix(self, ext):
-        assert ext.startswith('.')
-        name = self.name
-        if name == "":
-            raise ValueError(f"{self} has an empty name")
-        base, _ = os.path.splitext(name)
-        return self.parent / f"{base}{ext}"
+        stack = inspect.stack()
+        cname = stack[1].frame.f_locals["self"].__class__.__name__
```
kuenishi, line 82: I'm not sure what is intended here, regarding my local test.
```
>>> from pfio.v2 import pathlib, Local
>>> pathlib.Path('.', fs=Local())
Path('.')
>>> p = pathlib.Path('.', fs=Local())
>>> p.cwd
<bound method Path.cwd of <class 'pfio.v2.pathlib.Path'>>
>>> p.cwd()
  File "/home/kuenishi/mm/pfnet/pfio/pfio/v2/pathlib.py", line 373, in cwd
    raise _not_supported()
          ^^^^^^^^^^^^^^^^
  File "/home/kuenishi/mm/pfnet/pfio/pfio/v2/pathlib.py", line 82, in _not_supported
    cname = stack[1].frame.f_locals["self"].__class__.__name__
            ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'self'
```

pfio/v2/pathlib.py
```diff
@@ -1,271 +1,756 @@
...
+    @classmethod
+    def cwd(cls: Type[SelfPathType]) -> SelfPathType:
+        raise _not_supported()
 
-    def read_bytes(self):
-        with self.open(mode='rb') as fp:
-            return fp.read()
+    @classmethod
+    def home(cls: Type[SelfPathType]) -> SelfPathType:
+        raise _not_supported()
```
kuenishi, line 377: I agree handling this call as unsupported because pfio.v2.FS doesn't support it, but a naive call leads to this KeyError like in cwd.
```
>>> p.home()
  File "/home/kuenishi/mm/pfnet/pfio/pfio/v2/pathlib.py", line 377, in home
    raise _not_supported()
          ^^^^^^^^^^^^^^^^
  File "/home/kuenishi/mm/pfnet/pfio/pfio/v2/pathlib.py", line 82, in _not_supported
    cname = stack[1].frame.f_locals["self"].__class__.__name__
            ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'self
```

pfio/v2/pathlib.py
```diff
@@ -1,271 +1,756 @@
...
-    sep = '/'
-
-    def __init__(self, *args, fs=None, root=None):
-        '''
 
-        Empty argument indicates current directory in fs, fs.cwd
+@functools.lru_cache(maxsize=16)
+def _has_directory_feature(fs: FS) -> bool:
+    if isinstance(fs, S3):
+        # NOTE: Apache Ozone supports supports directory or folder.
```
kuenishi, line 27:
```suggestion
        # NOTE: Apache Ozone supports directories depending on its configuration.
```

pfio/v2/pathlib.py
```diff
@@ -1,271 +1,756 @@
...
+        `scheme` should moves to `FS`.
+    """
+
+    def __init__(
+        self,
+        *args: Union[str, PathLike],
+        fs: FS,
+        scheme: Optional[str] = None,
+    ) -> None:
+        if isinstance(fs, Local):
```
kuenishi, line 115: Q: do you @yhrkw  have a plan to support [custom scheme](https://pfio.readthedocs.io/en/latest/design.html#custom-scheme) in near future? Currently, `pfio.v2.FS` does not have `.scheme` property, but it'd be worth trying adding the feature to FS class and using it here.

kuenishi (changes requested):
Overall the implementation seems to be much better than original one. I appreceate you so much and want to merge this, given a few nitpicks addressed.

pfio/v2/pathlib.py
yhrkw: I fixed it and added `test_not_implemented` in `test_pathlib.py`.

pfio/v2/pathlib.py
yhrkw: Yes, I'm currently working on it. I will send a PR after merge.

kuenishi (approved):
Thank you for your contribution!

Issue #333, opened 2024-04-02
Support Python 3.12


kuenishi:
https://www.python.org/downloads/release/python-3120/

PR #334, opened 2024-04-10
Support python 3.12

kuenishi:
Addresses #333 .

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


kuenishi:
The CI is failing for sparse file cache, which is a deprecated feature.
```
=================================== FAILURES ===================================
______________________________ test_s3_zip[True] _______________________________

local_cache = True

    @mock_aws
    @pytest.mark.parametrize("local_cache", [False, True])
    def test_s3_zip(local_cache):
        with tempfile.TemporaryDirectory() as d:
```
So I decided to remove it beforehand at #335.

PR #337, opened 2024-04-16
Pick out document build at CI

kuenishi:
Trying to address CI failure: https://github.com/pfnet/pfio/pull/336#issuecomment-2060060780


kuenishi:
By default setup-python is going to use 3.10 so far and the doc build failed possibly due to this: https://github.com/pytest-dev/pytest/pull/9741

PR #338, opened 2024-04-17
add FS.scheme property

yhrkw:
ref: https://github.com/pfnet/pfio/pull/331#discussion_r1558905439

### DIFF
--- a/pfio/v2/fs.py
+++ b/pfio/v2/fs.py
@@ -87,9 +87,16 @@ class FS(abc.ABC):
     '''
 
     _cwd = ''
+    _scheme = ''
 
-    def __init__(self):
+    def __init__(self, scheme=None):
         self.pid = os.getpid()
+        if scheme:
+            self._scheme = str(scheme)
+
+    @property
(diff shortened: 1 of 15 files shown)


pfio/v2/fs.py
```diff
@@ -87,9 +87,16 @@ class FS(abc.ABC):
...
 
     _cwd = ''
+    _scheme = ''
 
-    def __init__(self):
+    def __init__(self, scheme=None):
         self.pid = os.getpid()
+        if scheme:
+            self._scheme = str(scheme)
+
```
kuenishi, line 96: I'd like you to add a few sentences of document here to clarify this variable stores the name of custom scheme or actual implementation scheme (e.g. s3, data, or hdfs). Reading the code told me either can be here. The former if custom scheme defined, and the latter if it's not defined.

kuenishi (changes requested):
The overall looks good, but I want a bit doc to clarify what the property is for.

namachan10777:
Duplicate with #377

Issue #258, opened 2022-01-11
Support PPE profiling


kuenishi:
One of requirements in our systems is to be suitable with other tools. As PPE has now Profiling feature, it would be nice if PFIO rides on its use cases, rather than inventing its own.

c.f. https://github.com/pfnet/pytorch-pfn-extras/pull/144

Update: PPE and PyTorch are both not ready for 3.10. https://github.com/pfnet/pytorch-pfn-extras/issues/386

Target tasks:
- [x] Local #326 / #340 
- [x] S3 #341 
- [x] Zip #344
- [x] CachedHttpFs #356 
- [x] FileCache / multiprocess FileCache #350 
- [x] HttpCache #356 
- [x] Add test suite with/without PPE

kuenishi:
Approach:
- Put PPE's decorator onto PFIO
- Check if we can get stats by adding current decorator
- Check the chrome tracing in outdir once PPE emit happened: c.f. https://github.com/pfnet/pytorch-pfn-extras/issues/729
- Save the start/end timing as trace with PPE

k5342:
All milestones are now completed 🎉
Thank you for working hard with us and your contribution, @KantaTamura!

kuenishi:
So it's now ready to ship :congratulations: . It'll be included in 2.9.0 release as the key feature of the new release.

PR #326, opened 2023-09-13
Profiling with PPE for Local

kuenishi:
PR draft to discuss the base design for https://github.com/pfnet/pfio/issues/258.

With a test script like below:
```python
import pytorch_pfn_extras as ppe

from pfio.v2 import Local, Path

@ppe.profiler.record_function("my_tag_2")
def my_run2(x):
    pass

def my_run1(x):
    pass

def p():
    with ppe.profiler.record("my_tag_1"):
        my_run1('hoge')

    for i in range(10):
        my_run2("far")

    total = 0
    count = 0
            
    p = Path()
    for f in p.glob("*"):
        s = f.stat()
        total += s.size
        count += 1

    d = 0
    with Local() as fs:
        for f in fs.list():
            if fs.isdir(f.strip()):
                d += 1
                continue
                
            len(fs.open(f).read())

    print(count - d, "files", d, "dirs", total, "bytes for files")

    ts = ppe.profiler.get_time_summary()

    with ts.summary() as o:
        print(o)

p()
```
It yields a result like this:
```
$ python p.py
15 files 13 dirs 8490 bytes for files
(<pytorch_pfn_extras.reporting.DictSummary object at 0x7fae6b3204d0>, {'my_tag_1.min': 9.5367431640625e-06, 'my_tag_1
.max': 9.5367431640625e-06, 'my_tag_2.min': 9.5367431640625e-07, 'my_tag_2.max': 2.86102294921875e-06, 'local-stat.mi
n': 9.059906005859375e-06, 'local-stat.max': 4.9591064453125e-05, 'local-open.min': 1.3828277587890625e-05, 'local-op
en.max': 6.914138793945312e-05, 'local-list.min': 6.9141387939453125e-06, 'local-list.max': 0.00014781951904296875})
```

PR #340, opened 2024-08-22
Support PPE profiling for `Local`

KantaTamura:
This PR resolves the `Local` task in #258 

> [!NOTE]
> To enable tracing, `trace=True` must be added during `Local()` initialization (or `from_url()`).

Sample program to check tracing
```python
import json
import pytorch_pfn_extras as ppe

from pfio.v2 import Local, Path, from_url

@ppe.profiler.record_function("my_tag_2", trace=True)
def my_run2(x):
    pass

def my_run1(x):
    pass

def p():
    tracer = ppe.profiler.get_tracer()

    tracer.clear()
    with ppe.profiler.record("my_tag_1", trace=True):
        my_run1('hoge')

    for i in range(4):
        my_run2("far")

    dir = 0
    fil = 0
    with Local(trace=True) as fs:
        for f in fs.list():
            if fs.isdir(f.strip()):
                dir += 1
                continue
            
            fil += 1
            len(fs.open(f).read())

    print(fil, "files", dir, "dirs")

    dict = tracer.state_dict()
    keys = [event["name"] for event in json.loads(dict["_event_list"])]
    print(keys)

    w = ppe.writing.SimpleWriter(out_dir="")

    # output '['
    tracer.initialize_writer("trace.json", w)
    # output json dump
    tracer.flush("trace.json", w)

p()
```

output
```
9 files 15 dirs
['my_tag_1', 'my_tag_2', 'my_tag_2', 'my_tag_2', 'my_tag_2', 'pfio.v2.Local:open', 'pfio.v2.Local:read', 'pfio.v2.Local:list-0', 'pfio.v2.Local:open', 'pfio.v2.Local:read', 'pfio.v2.Local:list-1', 'pfio.v2.Local:open', 'pfio.v2.Local:read', 'pfio.v2.Local:list-2', 'pfio.v2.Local:list-3', 'pfio.v2.Local:list-4', 'pfio.v2.Local:list-5', 'pfio.v2.Local:list-6', 'pfio.v2.Local:open', 'pfio.v2.Local:read', 'pfio.v2.Local:list-7', 'pfio.v2.Local:open', 'pfio.v2.Local:read', 'pfio.v2.Local:list-8', 'pfio.v2.Local:list-9', 'pfio.v2.Local:list-10', 'pfio.v2.Local:list-11', 'pfio.v2.Local:list-12', 'pfio.v2.Local:list-13', 'pfio.v2.Local:list-14', 'pfio.v2.Local:list-15', 'pfio.v2.Local:list-16', 'pfio.v2.Local:open', 'pfio.v2.Local:read', 'pfio.v2.Local:list-17', 'pfio.v2.Local:list-18', 'pfio.v2.Local:list-19', 'pfio.v2.Local:open', 'pfio.v2.Local:read', 'pfio.v2.Local:list-20', 'pfio.v2.Local:list-21', 'pfio.v2.Local:open', 'pfio.v2.Local:read', 'pfio.v2.Local:list-22', 'pfio.v2.Local:open', 'pfio.v2.Local:read', 'pfio.v2.Local:list-23']
```

drawing of the output json file (`trace.json`) with `chrome://tracing` is shown below
image

### DIFF
--- a/pfio/v2/local.py
+++ b/pfio/v2/local.py
@@ -5,18 +5,17 @@ import shutil
 from typing import Optional
 
 try:
-    from pytorch_pfn_extras.profiler import (  # type: ignore # NOQA
-        record_function, record_iterable)
+    from pytorch_pfn_extras.profiler import record_function, record_iterable
 
 except ImportError:
 
     # IF PPE is not available, wrap with noop
-    def record_function(*args):  # type: ignore # NOQA
+    def record_function(tag, trace, *args):  # type: ignore # NOQA
         def wrapper(f):
             return f
         return wrapper
 
-    def record_iterable(tag, iter, *args):   # type: ignore # NOQA
+    def record_iterable(tag, iter, trace, *args):   # type: ignore # NOQA
         yield from iter
 
 
@@ -99,7 +98,7 @@ class Local(FS):
             },
         )
 
-    @record_function("pfio.v2.Local:open")
+    @record_function("pfio.v2.Local:open", trace=True)
     def open(self, file_path, mode='r',
              buffering=-1, encoding=None, errors=None,
              newline=None, closefd=True, opener=None):
@@ -112,7 +111,8 @@ class Local(FS):
     def list(self, path: Optional[str] = '', recursive=False,
              detail=False):
         for e in record_iterable("pfio.v2.Local:list",
-                                 self._list(path, recursive, detail)):
+                                 self._list(path, recursive, detail),
+                                 trace=True):
             yield e
 
     def _list(self, path: Optional[str] = '', recursive=False,
@@ -151,7 +151,7 @@ class Local(FS):
                 yield from self._recursive_list(prefix_end_index,
                                                 e.path, detail)
 
-    @record_function("pfio.v2.Local:stat")
+    @record_function("pfio.v2.Local:stat", trace=True)
     def stat(self, path):
         path = os.path.join(self.cwd, path)
         return LocalFileStat(os.stat(path), path)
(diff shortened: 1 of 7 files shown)


k5342 (commented):
Overall LGTM. I have one minor suggestion.

k5342:
@kuenishi Could you take a look? Please let us know if you have any comments on the tag naming rule and the granularity of the tests.

setup.py
```diff
@@ -52,7 +52,7 @@
     packages=find_packages(),
     package_data={'pfio': package_data},
     extras_require={
-        'test': ['pytest', 'flake8', 'autopep8', 'parameterized', 'isort', 'moto[server]>=5.0.0', 'numpy', 'mypy'],
```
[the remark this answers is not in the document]
kuenishi: Agree. We'd rather add something like `trace` than adding the dependency to `test`. PFIO users shall be using profiling without this extra requires and will be installing PFIO with `pip install pfio` . In that case, PPE won't be installed even though they need it.

tests/v2_tests/test_local.py
```diff
@@ -291,6 +293,45 @@ def test_from_url_create_option(self):
         from_url(path, create=True)
         assert os.path.exists(path) and os.path.isdir(path)
 
+    def test_ppe_profiling(self):
```
kuenishi, line 296: Please don't extend the `TestLocal(unittest.TestCase)` class. We're slowly moving on from the `unitest` style, to a more pytest-like style where each test is defined as a single and independent function (sometimes with fixtures or parameterized). Please refer to `test_s3.py`.

kuenishi (changes requested):
The Python implementation itself looks good, but I have a few request to align with current style and options.

pfio/v2/local.py
```diff
@@ -4,9 +4,54 @@
 import shutil
 from typing import Optional
 
+try:
```
kuenishi, line 7: Refer to: https://github.com/pfnet/pfio/pull/326#discussion_r1329411782

KantaTamura: Split into `pfio/v2/_profiler.py` and changed to import from `_profiler.py`.

tests/v2_tests/test_local.py
KantaTamura: Changed to not use `TestLocal(unittest.TestCase)`.

Currently, `foo.txt` is generated in the directory where pytest runs, do you want to use a different path? (e.g. `/tmp/...`)

setup.py
```diff
@@ -52,11 +52,12 @@
...
+        'test': ['pytest', 'flake8', 'autopep8', 'parameterized', 'isort', 'moto[server]>=5.0.0', 'numpy', 'mypy', 'pytorch-pfn-extras'],
```
kuenishi, line 55: As we discussed offline, PPE shouldn't be here but tests must be switched in the test code using something like [skipif](https://docs.pytest.org/en/stable/how-to/skipping.html#id1).

KantaTamura: Changed to using `importorskip()` to skip if no ppe.

- no ppe
```
test_local.py ................s
```

- exist ppe
```
test_local.py .................
```

k5342: We may also need to tweak the tox.ini to enable CI testing for the tracing code, which we're adding this PR if we remove PPE from the test section in setup.py: https://github.com/pfnet/pfio/blob/b874d493abb80ca7836f439fb7af37bac327b182/tox.ini#L4-L20

kuenishi: Good catch

kuenishi (approved):
Please leave the summary of offline discussion with PPE developer why we don't need complete JSON file and it works fine without it.

kuenishi (changes requested):
Ah, and don't forget to update tox.ini to include `trace` option as @k5342 pointed out.

KantaTamura:
Offline discussion summary: Chrome Tracing does not require valid json and can be rendered with just `initialize_writer()` and `flush()` as in the [sample program](https://github.com/pfnet/pfio/pull/340#issue-2479646572).

ref. [Trace Event Format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#heading=h.yr4qxyxotyw)
> When provided as a string to the importer the ] at the end of the JSON Array Format is optional.

KantaTamura:
fix `tox.ini` and summarized the json output method in the documentation (tips).

kuenishi (approved):

PR #349, opened 2024-10-03
Revert "Settings for workload identity federation (gcs) (#346)"

k5342:
This reverts commit c4dff7cd3927cef40ae100787c5d1c60546711f5.

The CI invoked from external collaborators is failing, and we need to carefully redesign the CI and driver implementation, so we will revert the CI support for GCS once.


kmizumar (approved):
LGTM.

Issue #310, opened 2023-06-26
Drop Python 3.8 as of PEP-569


kuenishi:
Scheduled 2024/10/14
https://peps.python.org/pep-0569/

PR #354, opened 2024-10-17
Drop Python 3.8 as of PEP-569

k5342:
Addresses: https://github.com/pfnet/pfio/issues/310

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
 
     steps:
     - uses: actions/checkout@v4
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ learning training with batteries included. It supports
 ## Dependency
 
 - HDFS client and libhdfs for HDFS access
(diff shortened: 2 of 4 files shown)


k5342:
A grep result in my working directory. It seems to be dropped.
```
% grep 3\.8 -r ./ --exclude-dir=.mypy_cache --exclude-dir=.tox --exclude-dir=.git
./pfio/v2/zip.py:        # https://github.com/python/cpython/blob/3.8/Lib/zipfile.py#L392
grep: ./pfio/cache/__pycache__/multiprocess_file_cache.cpython-310.pyc: binary file matches
./pfio/cache/multiprocess_file_cache.py:    https://github.com/python/cpython/blob/3.8/Lib/tempfile.py#L406-L446
grep: ./docs/source/_static/classes_chainerio.png: binary file matches
grep: ./docs/source/_static/example1.png: binary file matches
grep: ./docs/source/_static/example2.png: binary file matches
./pfio.egg-info/PKG-INFO:Classifier: Programming Language :: Python :: 3.8
./pfio.egg-info/PKG-INFO:- CPython >= 3.8
./build/lib/pfio/v2/zip.py:        # https://github.com/python/cpython/blob/3.8/Lib/zipfile.py#L392
./build/lib/pfio/cache/multiprocess_file_cache.py:    https://github.com/python/cpython/blob/3.8/Lib/tempfile.py#L406-L446
```

kuenishi (approved):
:congratulations:

Issue #353, opened 2024-10-17
Support Python 3.13


kuenishi:

kuenishi:
Do we support GIL disabling? (That's the question) Or just ship as-is state?

k5342:
Support for the GIL disabling is pending for now because the `cryptography` package does not currently support it. This is used in the `moto` package, our dependency used in the testing.
- https://github.com/pyca/cryptography/issues/11436
```
moto==5.0.13
├── boto3 [required: >=1.9.201, installed: 1.34.133]
│   ├── botocore [required: >=1.34.133,<1.35.0, installed: 1.34.133]
│   │   ├── jmespath [required: >=0.7.1,<2.0.0, installed: 1.0.1]
│   │   ├── python-dateutil [required: >=2.1,<3.0.0, installed: 2.9.0.post0]
│   │   │   └── six [required: >=1.5, installed: 1.16.0]
│   │   └── urllib3 [required: >=1.25.4,<3,!=2.2.0, installed: 2.2.1]
│   ├── jmespath [required: >=0.7.1,<2.0.0, installed: 1.0.1]
│   └── s3transfer [required: >=0.10.0,<0.11.0, installed: 0.10.1]
│       └── botocore [required: >=1.33.2,<2.0a.0, installed: 1.34.133]
│           ├── jmespath [required: >=0.7.1,<2.0.0, installed: 1.0.1]
│           ├── python-dateutil [required: >=2.1,<3.0.0, installed: 2.9.0.post0]
│           │   └── six [required: >=1.5, installed: 1.16.0]
│           └── urllib3 [required: >=1.25.4,<3,!=2.2.0, installed: 2.2.1]
├── botocore [required: >=1.14.0, installed: 1.34.133]
│   ├── jmespath [required: >=0.7.1,<2.0.0, installed: 1.0.1]
│   ├── python-dateutil [required: >=2.1,<3.0.0, installed: 2.9.0.post0]
│   │   └── six [required: >=1.5, installed: 1.16.0]
│   └── urllib3 [required: >=1.25.4,<3,!=2.2.0, installed: 2.2.1]
├── cryptography [required: >=3.3.1, installed: 3.4.8]
├── Jinja2 [required: >=2.10.1, installed: 3.1.4]
│   └── MarkupSafe [required: >=2.0, installed: 2.1.5]
├── python-dateutil [required: >=2.1,<3.0.0, installed: 2.9.0.post0]
[...]

PR #355, opened 2024-10-17
Support Python 3.13

k5342:
Addresses: #353

Introduce Python 3.13 into CI and their environments, and tox.ini.


kuenishi:
It seems CI is failing only in 3.13.

kuenishi (approved):

PR #363, opened 2025-11-13
Print message to stderr

kuenishi:


k5342 (commented):
Do we write log message to stderr instead?

k5342 (approved):

Issue #359, opened 2024-10-24
Drop Python 3.9 at 2025-10 as of PEP596


kuenishi:
c.f. https://peps.python.org/pep-0596/

Once 3.9 dropped, we can officially say https://github.com/pfnet/pfio/issues/271 is somehow addressed as wontfix :P

namachan10777:
End-of-life support for Python 3.9 was completed by #364

PR #364, opened 2026-04-13
Python3.9 is EOL

namachan10777:
Due to the end-of-life for Python 3.9, we will remove support for Python 3.9 and eliminate any workarounds.

While version 3.14 has already been released, it contains several breaking changes that will require code modifications to address. We will handle 3.14 support in a separate pull request.

- Increase the minimum supported version to 3.10
- Remove Python 3.9 validation from CI
- Nativeize the `remove_prefix` function

## Related issues

- https://github.com/pfnet/pfio/issues/359

## User Impact

- `<=3.9`
  - makes installation impossible with some package managers


pfio/v2/pathlib.py
```diff
@@ -46,10 +46,7 @@ def _has_directory_feature(fs: FS) -> bool:
 
 
 def _removeprefix(text: str, prefix: str) -> str:
-    # NOTE: `str.removeprefix` supports version 3.9 or higher.
```
kuenishi, line 49: Well, I won't understand this change unless this comment, but I don't think this line is needed.

kuenishi (changes requested):
Tiny nit, but I want this fixed.

pfio/v2/pathlib.py
namachan10777: Indeed, it would be better to simply replace it with the default removeprefix function.

.github/workflows/python-package.yml
k5342:
```suggestion
        python-version: "3.14"
```

k5342 (commented):
CI might be flaky on the new environment. Could you please check the CI configuration? We also need tweak Python environment as well.

https://github.com/pfnet/pfio/pull/355

kuenishi:
Still failing on CI. Also, please address Sugihara-san's comment.

kuenishi:
If the master branch is already broken after long time silence, what about fixing the master before working on Python update?

namachan10777:
Currently, I'm performing environment maintenance including package management and testing with CI fixes. Please submit this PR and keep it in pending status until it gets merged.

kuenishi:
Let me know when this pull request gets ready for review.

namachan10777:
@kuenishi This PR is ready for review now.

kuenishi (approved):

Issue #365, opened 2026-04-14
Enhance CI stability


namachan10777:
The CI pipeline appears to be flaky. I want to improve both the stability of the CI process and reproducibility on my local machine by using uv and Docker.

## Related PRs

- https://github.com/pfnet/pfio/pull/364

PR #367, opened 2026-04-15
Fix linter warnings and test errors

namachan10777:
- remove `global _FORCE_LOCAL`
  - The `_FORCE_LOCAL` variable is read-only, making this line unnecessary.
- Handle `FileNotFoundError`
  - When `close()` explicitly deletes a file and then the garbage collector calls `close()` again, it attempts to unlink a non-existent file. This can trigger a `FileNotFoundError`, so Python 3.12 and later versions will generate a warning if this condition isn't properly handled.
- zipdir
  - This modification is only for testing purposes. Since zip files are created using relative paths from the current directory, multiple concurrent tests may cause conflicts that fail the tests. By creating temporary directories, we ensure that multiple tests running in parallel won't encounter conflicts, thereby improving test stability.

## Related issues/PRs

- https://github.com/pfnet/pfio/issues/365

### DIFF
--- a/pfio/cache/file_cache.py
+++ b/pfio/cache/file_cache.py
@@ -100,7 +100,6 @@ def _check_local(path: str):
     Raises an error if the ``stat(1)`` answers it as NFS.
 
     '''
-    global _FORCE_LOCAL
     if not _FORCE_LOCAL:
         return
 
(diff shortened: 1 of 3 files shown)


pfio/cache/file_cache.py
```diff
@@ -100,7 +100,6 @@ def _check_local(path: str):
     Raises an error if the ``stat(1)`` answers it as NFS.
 
     '''
-    global _FORCE_LOCAL
```
kuenishi, line 103: We don't need this global as this is only reference.

kuenishi (approved):
I'd like the PR title be capitalized, and not dots.

PR #368, opened 2026-04-17
Consolidate tool configs into pyproject.toml

namachan10777:
This migration moves tool configurations from setup.cfg and individual configuration files to pyproject.toml. Since only flake8 currently doesn't support pyproject.toml, this change will be implemented by extracting these configurations into separate files.

## Motivation

The pyproject.toml format has become widely adopted in modern Python tools, and we aim to make this transition for better interoperability with new tools.
This is part of the broader effort to update our development environment to newer standards.

## Related issues

- #365


kuenishi (approved):
Given that the test passes.

PR #370, opened 2026-04-20
Migrate CI and dependency management to uv

namachan10777:
This PR migrates dependency management, Python interpreter management, and virtual environment management to [uv](https://github.com/astral-sh/uv).
`uv` has rapidly gained traction in the Python community as an all-in-one tool for dependency, interpreter, and virtual environment management. Its extremely fast installation, the fact
  that it does not pollute the global environment, and its trajectory toward becoming a de facto standard make migration beneficial for pfio as well.

## Key benefits:

  - Unified dependency management
  - Simplified management of multiple Python interpreters
  - Reproducibility guaranteed by uv.lock

## Changes

  - Run tests via tox and generate documentation via sphinx through uv
    - tox uses tox-uv's uv-venv-lock-runner, so each per-Python env (and the doc env) is provisioned by uv sync --locked against uv.lock. Lockfile-backed reproducibility now covers the
  actual test/doc environments, not just a bootstrap venv
    - Local setup is a single uv run tox — dev dependencies are installed implicitly by uv run
  - Build the package with uv (uv build)
  - Unify the doc CI job into the test matrix so all tox envs share the same workflow definition

 ## Build artifact diff

 Only metadata-level changes are observed; no Python source files are modified in the resulting .whl or .tar.gz.

```diff
--- master/METADATA
+++ branch/METADATA
@@ -79 +79 @@
-$ make html
+$ uv run --extra doc make html
@@ -94,2 +94,2 @@
-$ pip install tox
-$ tox
+$ uv sync
+$ uv run tox --skip-env doc
```

```diff
--- master/RECORD
+++ branch/RECORD
@@ -21 +21 @@
-pfio-2.9.0.dist-info/METADATA,sha256=cSrO3dlLpJFZBnJqtG2DbDkD-KClN06EQZuZbV726Mc,3157
+pfio-2.9.0.dist-info/METADATA,sha256=ReXX3GfiCte6vKm2wbfyTNSR6TiYj6ITeb5D_4tXAlE,3190
```

```diff
--- master/README.md
+++ branch/README.md
@@ -32 +32 @@
-$ make html
+$ uv run --extra doc make html
@@ -47,2 +47,2 @@
-$ pip install tox
-$ tox
+$ uv sync
+$ uv run tox --skip-env doc
```
```diff
--- master/pyproject.toml
+++ branch/pyproject.toml
@@ -45,0 +46,3 @@
+[dependency-groups]
+dev = ["tox", "tox-uv"]
+
@@ -79,0 +83 @@
+runner = "uv-venv-lock-runner"
@@ -90,0 +95 @@
+runner = "uv-venv-lock-runner"
```


uv.lock
kuenishi, line 1: I don't think we should add a lock file because this is library package. We'd better even add it to .gitignore.

README.md
```diff
@@ -29,7 +29,7 @@ $ pip install .
 Documentation
 ```sh
 $ cd pfio/docs
-$ make html
+$ uv run --locked --extra doc make html
```
kuenishi: We shouldn't do any version locking on library development IMO.

kuenishi (changes requested):

namachan10777:
That's certainly correct. I've modified it to not use the lockfile.

kuenishi (approved):

Issue #372, opened 2026-04-23
Python 3.14 support


namachan10777:
The current pfio version cannot be used with 3.14 because it relies on changes or removed features introduced in 3.14. We will address these current 3.14 incompatibilities one by one before implementing full 3.14 support.

namachan10777:
- [x] #371 
- [x] #373 
- [x] #374 
- [x] #375 
- [x] Create PR for 3.14 support

kuenishi:
Please make sure that you mark at least one milestone for every issue you're going to work on.

namachan10777:
This is completed by #376

PR #374, opened 2026-04-24
Adapt PurePath.relative_to to Python 3.14

namachan10777:
Python 3.14 changes `pathlib.PurePath.relative_to` in two ways that break the pfio wrapper:

- The multi-arg form `relative_to(*multi)` was removed — stdlib raises TypeError. pfio's wrapper forwards `*other`, so any caller that passed multiple positional arguments (including our own tests) breaks.
- The new implementation skips `PurePosixPath` coercion for arguments that expose `with_segments`. pfio's `PurePath` defines `with_segments`, so passing a pfio Path to `relative_to` compares it against `self.parents` which remain stdlib paths; cross-type equality is always False and the subpath search raises ValueError (seen from `copy()`'s internal `entry.relative_to(src)`).

Unwrap pfio `PurePath` arguments to their inner `_pure` before delegating, so subpath checks compare like-typed values. For the removed multi-arg form, keep the public `*other` shape as a shim: join the parts into a single `PurePosixPath` and emit `DeprecationWarning` so downstream can migrate. A `NOTE:` in the wrapper flags the shim (and the `*other` signature) as future cleanup once callers have moved.

Update the existing pathlib test to use the single-arg form (the stdlib call in the same assertion would otherwise TypeError on 3.14), and add a dedicated test that exercises the shim and asserts the `DeprecationWarning`.

## Discussion

Allowing compatibility breaks would enable a simple solution of simply rejecting multiple arguments from the next pfio release.

## Related issues

- #372

### DIFF
--- a/pfio/v2/pathlib.py
+++ b/pfio/v2/pathlib.py
@@ -1,5 +1,6 @@
 import functools
 import uuid
+import warnings
 from fnmatch import fnmatch, fnmatchcase
 from io import IOBase
 from os import PathLike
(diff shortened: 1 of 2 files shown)


kuenishi (approved):
Looks good

